PR #7, opened 2021-11-15
added base test db config for docker-compose

Tom-Ritense:


Hildebrand-Ritense (approved):
I think this is exactly what we need to get the test environment working.

PR #43, opened 2021-12-14
Feature/32606 Store changes when document is edited

marijnritense:
Added a property "valtimo.audit.auditDocumentChanges" which, when set to true, audits document changes. Default is false.


Klaas-Ritense (approved):
This solution is not very pretty. But I don't have a better solution

PR #50, opened 2021-12-15
Feature/attachments

Tom-Ritense:
Fixes attachments sending. The html property needed to be empty string. After consulting with the helpdesk they spotted the issue.


mail/flowmailer/src/main/kotlin/com/ritense/mail/flowmailer/service/FlowmailerDispatcher.kt
```diff
@@ -57,14 +57,18 @@ class FlowmailerMailDispatcher(
     }
 
     override fun getMaximumSizeAttachments(): Int {
-        return MAX_SIZE_ATTACHMENTS
+        return MAX_SIZE_ATTACHMENTS_IN_BYTES
```
huibvanstraten, line 60: graag hernoemen naam MAX_SIZE EMAIL_BODY_IN_BYTES

huibvanstraten (approved):
That looks mighty fine

PR #66, opened 2021-12-30
Add support for PostgreSQL

ivo-ritense:


app/gzac/build.gradle
```diff
@@ -19,7 +19,9 @@ dependencies {
     implementation(project(":objects-api"))
 
     // Database
-    implementation "mysql:mysql-connector-java"
+    //implementation "mysql:mysql-connector-java"
```
marijnritense, line 22: How do we see this working in the future? We add both mysql and postgres?

app/gzac/src/main/resources/config/application.yml
```diff
@@ -45,7 +45,6 @@ spring:
         enabled: false
     jpa:
         database-platform: org.hibernate.dialect.MySQL57Dialect
-        database: mysql
```
marijnritense, line 48: I'm guessing this is derived from database-platform instead?

audit/src/main/resources/config/liquibase/changelog/20210414-audit-record-add-document-id-changelog.xml
```diff
@@ -19,11 +19,27 @@
...
+            <dbms type="mysql"/>
+        </preConditions>
         <addColumn tableName="audit_record">
             <column name="document_id" type="BINARY(16)"/>
         </addColumn>
     </changeSet>
+
+    <changeSet author="ivozaal" id="audit-record-add-document-id-no-mysql">
+        <preConditions onFail="MARK_RAN">
+            <dbms type="postgresql"/>
```
theo-ritense, line 32: This is not really a "no-mysql" changeSet, this is a "postgresql" changeSet. Not sure if the hash changes if you want to add another dbms in the future.

You should be able to do a real '!mysql' by using:
`<not><dbms type="mysql"/></not>`

audit/src/main/java/com/ritense/audit/autoconfigure/AuditAutoConfiguration.java
```diff
@@ -76,4 +84,18 @@ public AuditEventListener auditEventListener(AuditEventProcessor auditEventProce
         return new AuditEventListenerImpl(auditEventProcessor);
     }
 
+    @Bean
+    @ConditionalOnProperty(prefix = "valtimo", name = "database", havingValue = "postgres")
+    public JpaRepositoryFactoryBean<AuditRecordRepository<AuditRecord, AuditRecordId>, AuditRecord, AuditRecordId> postgresAuditRecordRepository() {
```
Tom-Ritense, line 89: Waar is die JpaRepositoryFactoryBean voor nodig?

theo-ritense (changes requested):
Please check my comment about "no-mysql"

audit/src/main/resources/config/liquibase/changelog/20210414-audit-record-add-document-id-changelog.xml
marijnritense: Preconditions are not used in the checksum. I'm more in favor of whitelisting rather than blacklisting, so making this specific to postgresql for now is fine. Other dbms can be added in the future

theo-ritense (approved):

audit/src/main/java/com/ritense/audit/autoconfigure/AuditAutoConfiguration.java
ivo-ritense: Dat is de manier om in spring een implementatie van een repository interface op te geven. Je kan niet `new SomethingRepository()` doen omdat het een interface is. Zo kan je dus kiezen wel DBMS specifieke repository er gebruikt wordt.

app/gzac/src/main/resources/config/application.yml
ivo-ritense: Volgens spring:

Target database to operate on, auto-detected by default. Can be alternatively set using the "databasePlatform" property.

audit/src/main/resources/config/liquibase/changelog/20190305-create-audit-record-changelog.xml
```diff
@@ -37,23 +40,123 @@
...
     <changeSet author="tom (generated)" id="1551802246513-5">
+        <preConditions onFail="MARK_RAN">
+            <dbms type="mysql"/>
+        </preConditions>
         <createIndex indexName="user_index" tableName="audit_record">
             <column name="user"/>
         </createIndex>
     </changeSet>
+
+    <changeSet author="Ivo Zaal" id="1551802246513-6">
```
marijnritense, line 75: These should run without issues, but I do want to point out we should try to never add new changesets to existing changelogs for already supported DBMSs. These should be in a separate changelog.

connector/src/main/resources/config/liquibase/changelog/20210804-connector-changelog.xml
```diff
@@ -43,15 +50,93 @@
...
+        <preConditions onFail="MARK_RAN">
+            <dbms type="mysql"/>
+        </preConditions>
         <createIndex indexName="ct_name_i" tableName="connector_instance">
             <column name="name"/>
         </createIndex>
     </changeSet>
 
+    <!-- postgres does't support using the same index name so we have to rename them for mysql instances -->
+    <changeSet author="Ivo Zaal" id="1628085538996-5">
```
marijnritense, line 73: Same here, new changesets for existing DBMSs should go into a new changelog.

audit/src/main/resources/config/liquibase/changelog/20190305-create-audit-record-changelog.xml
ivo-ritense: No sure. i feel like it would be more clear for future development if we group the changesets for the same purpose together.

Rick-Ritense (approved):

audit/src/main/resources/config/liquibase/changelog/20190305-create-audit-record-changelog.xml
marijnritense: Isn't that the opposite of what you want? By adding changesets to an existing changelog, you create a difference in how changesets are executed for already existing projects with up and running production environments vs new projects. I would group it based on release. For the postgres changesets, these obviously have to happen when the original mysql changesets were planned, but anything else should be in a 9.3.0 or 9.4.0 changelog, or whatever version the changes are for.

ivo-ritense: These changesets are meant to make sure that the mysql and postgres index names are the same though. If we put them in a new file, where are the postgres changesets meant to go? they are not completely equivalent

marijnritense: The postgres changes are all necessary, so I would leave them here and only move the mysql changes instead.

marijnritense (approved):

app/gzac/build.gradle
ivo-ritense: I think we just want to include both database drivers and have the spring configuration/profile decide which one gets used.

audit/src/main/resources/config/liquibase/changelog/20190305-create-audit-record-changelog.xml
ivo-ritense: I'm still not sure i completely agree, but i moved the changesets to a new changelog

PR #72, opened 2022-01-12
#32838 aantekening (van klantcontact) vastleggen buiten proces

Klaas-Ritense:


contactmoment/src/main/kotlin/com/ritense/contactmoment/web/rest/request/CreateContactMomentRequest.kt
```diff
@@ -0,0 +1,22 @@
...
+ * distributed under the License is distributed on an "AS IS" basis,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.ritense.contactmoment.web.rest.request
+
+data class CreateContactMomentRequest(
+    val kanaal: String,
```
ivo-ritense, line 20: I think the kanaal wasn't part of the story. If we want to do this i think it should be an enum. Something like: Phone, Mail, Other

Klaas-Ritense: I removed kanaal from the request for now. we can always add it later.
And made kanaal enum

Rick-Ritense (approved):

PR #75, opened 2022-01-13
feature/build excel export 

huibvanstraten:
- parsing data to Excel file 
- Handles different types of variables
- in case of large quantities of data, rows can be appended to the created excel file
- function runs asynchronous


huibvanstraten:
Project is on hold till further notice.

PR #83, opened 2022-01-28
#33180 koppeling smartdocuments basis

Klaas-Ritense:
Screenshot 2022-01-28 at 09 09 09
Screenshot 2022-01-28 at 09 08 27
Screenshot 2022-01-28 at 09 08 59


document-generation/smartdocuments/src/main/kotlin/com/ritense/smartdocuments/service/CamundaSmartDocumentGenerator.kt
```diff
@@ -0,0 +1,92 @@
...
+        val camundaPropertiesMap = mutableMapOf<String, Any>()
+        execution
+            .bpmnModelElementInstance
+            .extensionElements
+            .elementsQuery
+            .filterByType(CamundaProperties::class.java)
+            .singleResult()
+            .camundaProperties
+            .associateTo(camundaPropertiesMap) {
+                it.camundaName to getPlaceholderValue(it.camundaValue, execution, document)
```
Rick-Ritense, line 64: This could result in the same NullPointerException that was part of the red flag stories. It is possible to set a property without setting either the name or the value. If either or both of them are not filled a NPE is thrown.

I created a new story for this problem but maybe you can already fix it at least for this case

Rick-Ritense (approved):

document-generation/smartdocuments/src/main/kotlin/com/ritense/smartdocuments/service/CamundaSmartDocumentGenerator.kt
```diff
@@ -0,0 +1,89 @@
...
+            .elementsQuery
+            .filterByType(CamundaProperties::class.java)
+            .singleResult()
+            .camundaProperties
+            .filter { it.camundaName != null && it.camundaValue != null }
+            .associate { it.camundaName!! to getPlaceholderValue(it.camundaValue, execution, document) }
+    }
+
+    private fun getPlaceholderValue(value: String, execution: DelegateExecution, document: Document): Any {
+        return if (value.startsWith("pv:")) {
```
ivo-ritense, line 67: I'd rather not hardcode this. Is it possible to use the FormFieldDataResolvers to find this?

PR #100, opened 2022-02-21
Feature/33595 document definition roles

theo-ritense:


document/src/main/resources/config/liquibase/changelog/20220215-create-document-definition-role-changelog.xml
```diff
@@ -0,0 +1,44 @@
+<?xml version="1.1" encoding="UTF-8" standalone="no"?>
```
Klaas-Ritense, line 1: Existing project that migrate to this version, will they keep seeing their cases?

Do we need to add a migration script that will make sure that all users of existing projects will keep seeing all cases?

Klaas-Ritense (changes requested):

document/src/main/resources/config/liquibase/changelog/20220215-create-document-definition-role-changelog.xml
Rick-Ritense: They will always see them as an administrator.
But no by default all the cases will be disabled for all roles. So the admin must set the correct roles before they will be visible for a ROLE_USER

document/src/main/java/com/ritense/document/domain/DocumentDefinitionRole.java
```diff
@@ -0,0 +1,40 @@
+/*
+ * Copyright 2015-2020 Ritense BV, the Netherlands.
```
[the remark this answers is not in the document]
theo-ritense: I guess we should update the whole project then. Not part of this story.

document/src/main/java/com/ritense/document/autoconfigure/DocumentAutoConfiguration.java
```diff
@@ -79,11 +81,13 @@ public JsonSchemaDocumentService documentService(
...
         final ResourceLoader resourceLoader,
-        final DocumentDefinitionRepository<JsonSchemaDocumentDefinition> documentDefinitionRepository
+        final DocumentDefinitionRepository<JsonSchemaDocumentDefinition> documentDefinitionRepository,
+        final DocumentDefinitionRoleRepository<JsonSchemaDocumentDefinitionRole> documentDefinitionRoleRepository
     ) {
         return new JsonSchemaDocumentDefinitionService(
             resourceLoader,
-            documentDefinitionRepository
+            documentDefinitionRepository,
+            documentDefinitionRoleRepository
```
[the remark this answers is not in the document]
theo-ritense: Had a discussion with @ThomasRitense about this. I believe there's no way to implement this cleanly without having a breaking change. Plus: the repositories are autowired and should not be breaking on a standard configuration.

document/src/main/resources/config/liquibase/changelog/20220215-create-document-definition-role-changelog.xml
theo-ritense: I added a migration script for existing installations, where ROLE_USER is assigned to the document definitions.
Any new definitions added during or after the upgrade will not have any role assigned by default.

Klaas-Ritense (approved):

PR #105, opened 2022-02-22
Fix tests

Klaas-Ritense:


Rick-Ritense (approved):

klant/src/test/resources/application.yml
```diff
@@ -0,0 +1,3 @@
+valtimo:
+    openklant:
+        url: "http://localhost:8006"
```
ivo-ritense, line 3: Why would you need this on a specific port for a test? is it to trick the system into creating the klantservices because of the ConditionalOnProperty?

Klaas-Ritense: The url can be anything. Its just to fill the ConditionalOnProperty. Otherwise we get a NoSuchBeanDefinitionException.

PR #142, opened 2022-03-23
34491 support kvk for productaanvraag

ivo-ritense:


klant/src/main/kotlin/com/ritense/klant/service/impl/OpenKlantService.kt
```diff
@@ -0,0 +1,32 @@
...
+    }
+
+    protected fun getDefaultWebsiteUrl(): String {
+        return "http://example.org"
+    }
+
+    private fun validateKlantnummerNotTaken(klantnummer: String): Boolean {
+        val klantPage = openKlantClient.searchKlanten(
+            KlantSearchFilter(
+            klantnummer = klantnummer
```
marijnritense, line 27: Funky indentation, think this and the next line need 1 more.

marijnritense (approved):

PR #147, opened 2022-03-28
#34289 autodeployment form flow definities

Klaas-Ritense:


form-flow/src/main/kotlin/com/ritense/formflow/service/Mapper.kt
```diff
@@ -0,0 +1,36 @@
...
+
+    private val mapper = ObjectMapper()
+
+    fun get(): ObjectMapper {
+        return mapper
+    }
+
+    init {
+        mapper.registerModule(KotlinModule())
+            .configure(DeserializationFeature.FAIL_ON_UNKNOWN_PROPERTIES, false)
```
marijnritense, line 33: Should fail, right? If not all properties are found in the class, then de-serialization should fail. The opposite, however, should not be true. Is there a reason to have this be false?

form-flow/src/main/kotlin/com/ritense/formflow/service/FormFlowDeploymentService.kt
```diff
@@ -0,0 +1,116 @@
...
+
+    fun deploy(formFlowKey: String, formFlowJson: InputStream) {
+        deploy(formFlowKey, StreamUtils.copyToString(formFlowJson, StandardCharsets.UTF_8))
+    }
+
+    fun deploy(formFlowKey: String, formFlowJson: String) {
+        validate(formFlowKey, formFlowJson)
+
+        val formFlowDefinition = Mapper.get().readValue(formFlowJson, FormFlowDefinition::class.java)
+        formFlowDefinition.steps.forEach { step -> step.id.formFlowDefinition = formFlowDefinition }
```
marijnritense, line 68: I think this needs to happen in the init of the FormFlowDefinition (so it is longer the responsibility of the AutoDeployer). I don't know if this will work with the mapper, but if it does not, then we need a more specific mapper (rather than the generated one).

form-flow/src/main/kotlin/com/ritense/formflow/repository/FormFlowDefinitionRepository.kt
```diff
@@ -20,4 +20,6 @@ import com.ritense.formflow.domain.FormFlowDefinition
 import com.ritense.formflow.domain.FormFlowDefinitionId
 import org.springframework.data.jpa.repository.JpaRepository
 
-interface FormFlowDefinitionRepository : JpaRepository<FormFlowDefinition, FormFlowDefinitionId>
\ No newline at end of file
+interface FormFlowDefinitionRepository : JpaRepository<FormFlowDefinition, FormFlowDefinitionId> {
```
marijnritense, line 23: Should this class be public? I think package-private makes more sense.

form-flow/src/main/kotlin/com/ritense/formflow/service/FormFlowDeploymentService.kt
```diff
@@ -0,0 +1,116 @@
...
+
+            formFlowDefinition.id = definitionId
+            formFlowService.save(formFlowDefinition)
+            logger.info("Deployed Form Flow - {}", definitionId.toString())
+        } catch (e: Exception) {
+            throw RuntimeException("Failed to deploy Form Flow $formFlowKey", e)
+        }
+    }
+
+    private fun validate(formFlowKey: String, formFlowJson: String) {
```
marijnritense, line 92: Why is the key necessary if it's already in the formFlowJson?

form-flow/src/main/kotlin/com/ritense/formflow/service/FormFlowDeploymentService.kt
```diff
@@ -0,0 +1,116 @@
...
+        val definitionJsonObject = JSONObject(JSONTokener(formFlowJson))
+        if (formFlowKey != definitionJsonObject.get("key")) {
```
marijnritense, line 94: This statement can be removed.

form-flow/src/main/kotlin/com/ritense/formflow/service/FormFlowDeploymentService.kt
```diff
@@ -0,0 +1,116 @@
...
+            throw RuntimeException("Form Flow file name '$formFlowKey' doesn't match key property '${definitionJsonObject.get("key")}'")
+        }
+        if (definitionJsonObject.toString().isEmpty()) {
```
marijnritense, line 97: This is a bit confusing, what's this for?

form-flow/src/main/kotlin/com/ritense/formflow/service/FormFlowService.kt
```diff
@@ -0,0 +1,37 @@
...
+
+    fun findLatestDefinitionByKey(formFlowKey: String): FormFlowDefinition? {
+        return formFlowDefinitionRepository.findFirstByIdKeyOrderByIdVersionDesc(formFlowKey)
+    }
+
+    fun save(formFlowDefinition: FormFlowDefinition) {
+        formFlowDefinitionRepository.findById(formFlowDefinition.id).ifPresentOrElse({
+            throw UnsupportedOperationException("Failed to save From Flow. Form Flow already exists: ${formFlowDefinition.id}")
+        }, {
+            formFlowDefinitionRepository.saveAndFlush(formFlowDefinition)
```
marijnritense, line 34: Why a saveAndFlush?

form-flow/src/main/kotlin/com/ritense/formflow/service/FormFlowService.kt
```diff
@@ -0,0 +1,37 @@
...
+
+class FormFlowService(
+    private val formFlowDefinitionRepository: FormFlowDefinitionRepository
+) {
+
+    fun findLatestDefinitionByKey(formFlowKey: String): FormFlowDefinition? {
+        return formFlowDefinitionRepository.findFirstByIdKeyOrderByIdVersionDesc(formFlowKey)
+    }
+
+    fun save(formFlowDefinition: FormFlowDefinition) {
```
marijnritense, line 30: I'd like to move this to the FormFlowDefinition in my own PR. No need to do anything with this right now, but I want to show some solutions that I think can be a lot cleaner to work with.

marijnritense (changes requested):

form-flow/src/test/kotlin/com/ritense/formflow/service/FormFlowDeploymentServiceIntTest.kt
```diff
@@ -0,0 +1,86 @@
...
+    lateinit var formFlowStepRepository: FormFlowStepRepository
+
+    @BeforeEach
+    fun beforeEach() {
+        formFlowDeploymentService.deployAll()
+    }
+
+    @AfterEach
+    fun afterEach() {
+        formFlowStepRepository.deleteAll()
```
ivo-ritense, line 49: Doesn't transactional work instead?

form-flow/src/test/kotlin/com/ritense/formflow/service/FormFlowDeploymentServiceIntTest.kt
```diff
@@ -0,0 +1,86 @@
...
+        assertThat(inkomensLoket.id.key).isEqualTo("inkomens_loket")
+        assertThat(inkomensLoket.id.version).isEqualTo(1L)
+        assertThat(inkomensLoket.startStep).isEqualTo("woonplaats")
+        assertThat(inkomensLoket.steps).hasSize(7)
+    }
+
+    @Test
+    fun `should not deploy same Form Flow twice`() {
+        formFlowDeploymentService.deployAll()
+        formFlowDeploymentService.deployAll()
```
ivo-ritense, line 68: The second deployAll() isn't needed here because it's already done in the beforeEach() It might be better to make it more consistent by adding the deploy to the previous test, or removing 1 deployAll in this method.

form-flow/src/test/resources/config/form-flow/inkomens_loket.json
```diff
@@ -0,0 +1,62 @@
...
+
+        {
+            "key": "leeftijd",
+            "nextSteps": [
+                {
+                    "condition": "${submissionData.leeftijd.isOuderDan21 == true && submissionData.leeftijd.isJongerDanAOW == true}",
+                    "step": "gezinssituatie"
+                },
+                {
+                    "step": "end"
```
ivo-ritense, line 27: Not sure if it's on purpose, but i like how the condition is missing here, and set explicitly to null in the step above. This way both configuration methods are tested.

form-flow/src/main/resources/config/form-flow/schema/formflow.schema.json
```diff
@@ -0,0 +1,72 @@
...
+                                "step": {
+                                    "type": "string",
+                                    "title": "The name of the next step. This should match one of the keys in the steps array"
+                                },
+                                "condition": {
+                                    "type": [
+                                        "string",
+                                        "null"
+                                    ],
+                                    "title": "This should contain an expression telling what condition should be met to go to this next step"
```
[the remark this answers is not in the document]
Klaas-Ritense: Also added it here

form-flow/src/main/resources/config/form-flow/schema/formflow.schema.json
```diff
@@ -0,0 +1,72 @@
...
+                                }
+                            },
+                            "required": [
+                                "step"
+                            ]
+                        }
+                    },
+                    "nextStep:": {
+                        "type": "string",
+                        "description": "The name of the next step in the flow. When multiple steps should be available you should use the property 'nextSteps'"
```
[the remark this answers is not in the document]
Klaas-Ritense: changed

form-flow/src/test/kotlin/com/ritense/formflow/service/FormFlowDeploymentServiceIntTest.kt
Klaas-Ritense: changed it

form-flow/src/test/kotlin/com/ritense/formflow/service/FormFlowDeploymentServiceIntTest.kt
Klaas-Ritense: Yes, changed it

form-flow/src/test/resources/config/form-flow/inkomens_loket.json
Klaas-Ritense: Yes, was on purpose

form-flow/src/main/kotlin/com/ritense/formflow/autoconfigure/FormFlowAutoConfiguration.kt
```diff
@@ -18,11 +18,34 @@ package com.ritense.formflow.autoconfigure
...
+import com.ritense.formflow.service.FormFlowService
+import org.springframework.boot.autoconfigure.condition.ConditionalOnMissingBean
 import org.springframework.boot.autoconfigure.domain.EntityScan
+import org.springframework.context.annotation.Bean
 import org.springframework.context.annotation.Configuration
+import org.springframework.core.io.ResourceLoader
 import org.springframework.data.jpa.repository.config.EnableJpaRepositories
 
 @Configuration
 @EnableJpaRepositories(basePackageClasses = [FormFlowDefinitionRepository::class, FormFlowStepRepository::class])
```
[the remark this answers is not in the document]
marijnritense: Okay, then let's only put 1 class in here. What this does is it takes the package based on the class, so putting down two means you're defining the same base package twice.

form-flow/src/main/kotlin/com/ritense/formflow/domain/FormFlowDefinition.kt
```diff
@@ -30,12 +31,29 @@ import javax.persistence.Table
 data class FormFlowDefinition(
 
     @EmbeddedId
-    val id: FormFlowDefinitionId,
+    @JsonProperty("key")
```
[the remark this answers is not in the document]
Klaas-Ritense: Not sure what you mean here

`Does this mean the version number is mandatory in the definition`
The version is not mandatory

`If you remove the "data" from this class`
What do you mean by "data"

`you can accept key: String as an argument.`
The constructor of FormFlowDefinitionId is already accepting a key String as argument

form-flow/src/main/kotlin/com/ritense/formflow/service/FormFlowDeploymentService.kt
marijnritense: I get that it's done for document definitions as well, but it seems like such an arbitrary restriction. What's the point of making sure the filename is the same as the key?

form-flow/src/main/kotlin/com/ritense/formflow/service/FormFlowDeploymentService.kt
marijnritense: Let's use filename for the key.

marijnritense (approved):

ivo-ritense (approved):

PR #163, opened 2022-04-07
Reverse proxy support for OpenZaak resource service

Hildebrand-Ritense:
Added reverse proxy support in OpenZaak resource service, serving an HTTPS URL in the response if the source requests was reverse proxies with x-forwarded-proto: https


theo-ritense (changes requested):
I believe this is not necessary, because Spring Boot can be configured to handle this for you:
https://exchangetuts.com/how-to-cope-with-x-forwarded-headers-in-spring-boot-220-spring-web-mvc-behind-reverse-proxy-1639837101295022 

If it works the same as the x-forwarded-for which changes the request.getRemoteAddress() output, you should not need to change any code.

Hildebrand-Ritense:
You are right. In the official docs as well: https://docs.spring.io/spring-boot/docs/current/reference/html/howto.html#howto.webserver.use-behind-a-proxy-server
Fixed this using server.forward-headers-strategy: native property.

PR #164, opened 2022-04-08
#34328 Evalueren expressies in onOpen property

Klaas-Ritense:
`formFlowService.open(...)`

### DIFF
--- a/form-flow/src/main/kotlin/com/ritense/formflow/expression/SpelExpressionProcessor.kt
+++ b/form-flow/src/main/kotlin/com/ritense/formflow/expression/SpelExpressionProcessor.kt
@@ -22,11 +22,12 @@ import org.springframework.expression.EvaluationContext
 import org.springframework.expression.Expression
 import org.springframework.expression.ParseException
 import org.springframework.expression.ParserContext
+import org.springframework.expression.common.TemplateParserContext
 import org.springframework.expression.spel.standard.SpelExpressionParser
 import org.springframework.expression.spel.support.StandardEvaluationContext
 
 class SpelExpressionProcessor(
-    private val parserContext: ParserContext = ParserContext.TEMPLATE_EXPRESSION,
+    private val parserContext: ParserContext = TemplateParserContext("\${", "}"),
     private val evaluationContext: EvaluationContext = StandardEvaluationContext()
 ) {
 
--- a/form-flow/src/main/kotlin/com/ritense/formflow/expression/SpelExpressionProcessor.kt
+++ b/form-flow/src/main/kotlin/com/ritense/formflow/expression/SpelExpressionProcessor.kt
@@ -1,0 +1,60 @@
+/*
+ * Copyright 2015-2022 Ritense BV, the Netherlands.
+ *
+ * Licensed under EUPL, Version 1.2 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * https://joinup.ec.europa.eu/collection/eupl/eupl-text-eupl-12
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" basis,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.ritense.formflow.expression
+
+import com.ritense.formflow.exception.ExpressionExecutionException
+import com.ritense.formflow.exception.ExpressionParseException
+import org.springframework.expression.EvaluationContext
+import org.springframework.expression.Expression
+import org.springframework.expression.ParseException
+import org.springframework.expression.ParserContext
+import org.springframework.expression.spel.standard.SpelExpressionParser
+import org.springframework.expression.spel.support.StandardEvaluationContext
+
+class SpelExpressionProcessor(
+    private val parserContext: ParserContext = ParserContext.TEMPLATE_EXPRESSION,
+    private val evaluationContext: EvaluationContext = StandardEvaluationContext()
+) {
+
+    fun <T> process(expression: String, resultType: Class<T>? = null): T? {
+        val spelExpression = parseSpelExpression(expression)
+
+        return if (spelExpression != null) {
+            try {
+                spelExpression.getValue(evaluationContext, resultType)
+            } catch (e: RuntimeException) {
+                throw ExpressionExecutionException(expression, e)
+            }
+        } else {
+            null
+        }
+    }
+
+    fun validate(expression: String) {
+        parseSpelExpression(expression)
+    }
+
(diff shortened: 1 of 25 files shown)


form-flow/src/main/kotlin/com/ritense/formflow/service/FormFlowDeploymentService.kt
```diff
@@ -98,6 +96,14 @@ class FormFlowDeploymentService(
         schema.validate(definitionJsonObject)
     }
 
+    private fun validate(formFlowDefinitionConfig: FormFlowDefinition) {
```
marijnritense, line 99: Shouldn't this validate be in the config class itself?

form-flow/src/main/kotlin/com/ritense/formflow/service/FormFlowService.kt
```diff
@@ -36,6 +42,36 @@ class FormFlowService(
         return formFlowDefinitionRepository.getById(formFlowDefinitionId)
     }
 
+    fun open(formFlowInstance: FormFlowInstance) {
```
marijnritense, line 45: This should probably be a either a helper class, or inside of the (definition) step itself, so the instance can ask the definition how onOpen should be resolved. Regardless, the open() should be performed in the instance, and not in the FormFlowService.

form-flow/src/main/kotlin/com/ritense/formflow/service/FormFlowService.kt
```diff
@@ -36,6 +42,36 @@ class FormFlowService(
...
+                        formFlowInstance.formFlowDefinition.id.key,
+                        expression,
+                        e
+                    )
+                }
+            }
+        }
+    }
+
+    fun parseSpelExpression(expression: String, context: ParserContext? = null): Expression? {
```
marijnritense, line 65: I think this needs to move to the definition. Probably on FormFlowStep level.

marijnritense (changes requested):

form-flow/src/main/kotlin/com/ritense/formflow/service/FormFlowService.kt
Klaas-Ritense: How will implementation projects be able to extend or change the 'open' logic, if its in the definition

form-flow/src/main/kotlin/com/ritense/formflow/service/FormFlowService.kt
Klaas-Ritense: Both FormFlowStep and FormFlowNextStep are going to use this and perhaps even more in the future. Since it doesn't tie to one specific domain entity, its better to put it into a service

form-flow/src/main/kotlin/com/ritense/formflow/service/FormFlowService.kt
```diff
@@ -36,6 +42,36 @@ class FormFlowService(
...
+        val expressionMatcher = "^\\\$\\{(.*)\\}\$".toRegex().find(expression) ?: return null
```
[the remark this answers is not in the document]
Klaas-Ritense: Ja, dat is overzichtelijker. Ik weet alleen nog niet hoe de expressies er uit moeten komen te zien: `${expressie}` of `#{expressie}` of `expressie`

form-flow/src/main/kotlin/com/ritense/formflow/service/FormFlowService.kt
marijnritense: They don't. They're not supposed to.

form-flow/src/main/kotlin/com/ritense/formflow/service/FormFlowService.kt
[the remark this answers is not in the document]
marijnritense: `${expressie}`

form-flow/src/main/kotlin/com/ritense/formflow/service/FormFlowService.kt
marijnritense: I disagree with putting it in the service, as this is logic that does not require injection of any sort. I see two options:

1. Add a static method to a helper class. Not a huge fan of this, but it is an option.
2. Create a base class that FormFlowStep and FormFlowNextStep both extend, so the method is available to both of them.

form-flow/src/main/kotlin/com/ritense/formflow/service/FormFlowService.kt
Klaas-Ritense: I looked it up. There is an interesting talk about this exact topic: https://www.youtube.com/watch?v=US8QG9I1XW0&t=1994s

```
The more behavior you add to the class, the more behavior it also attracts. So you'd end up with a couple of very big classes ... The classes attract more and more behavior and become less and less cohesive. We end up calling those classes 'god classes'. That is a well-known anti pattern.
```

form-flow/src/main/kotlin/com/ritense/formflow/expression/SpelExpressionProcessor.kt
```diff
@@ -0,0 +1,60 @@
...
+import com.ritense.formflow.exception.ExpressionParseException
+import org.springframework.expression.EvaluationContext
+import org.springframework.expression.Expression
+import org.springframework.expression.ParseException
+import org.springframework.expression.ParserContext
+import org.springframework.expression.spel.standard.SpelExpressionParser
+import org.springframework.expression.spel.support.StandardEvaluationContext
+
+class SpelExpressionProcessor(
+    private val parserContext: ParserContext = ParserContext.TEMPLATE_EXPRESSION,
```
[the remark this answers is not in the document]
theo-ritense: Fixed

form-flow/src/main/kotlin/com/ritense/formflow/service/FormFlowService.kt
```diff
@@ -36,6 +38,18 @@ class FormFlowService(
         return formFlowDefinitionRepository.getById(formFlowDefinitionId)
     }
 
+    fun open(formFlowInstanceId: FormFlowInstanceId) {
+        val formFlowInstance = formFlowInstanceRepository.getById(formFlowInstanceId)
+        val currentStep = formFlowInstance.getCurrentStep()
+        val stepDefinition = currentStep.definition
+
+        val expressionProcessor = expressionProcessorFactory.create(currentStep)
```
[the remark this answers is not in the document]
theo-ritense: The current step holds a reference to the formFlowInstance, which in turn has the history.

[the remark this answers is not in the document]
marijnritense: Fair, Then it makes sense to have a method in FormFlowStepInstance that can retrieve the (relevant) history, rather than exposing the entirety of the instance (and as a result things like definition too).

marijnritense (approved):

PR #165, opened 2022-04-12
Feature/34623 step on complete

theo-ritense:


form-flow/src/main/kotlin/com/ritense/formflow/service/FormFlowService.kt
```diff
@@ -39,13 +40,20 @@ class FormFlowService(
...
+    fun complete(formFlowInstanceId: FormFlowInstanceId) {
+        processExpressions(formFlowInstanceId, FormFlowStep::onComplete)
+    }
+
+    private fun processExpressions(formFlowInstanceId: FormFlowInstanceId, expressionist: (FormFlowStep)-> List<String>?) {
         val formFlowInstance = formFlowInstanceRepository.getById(formFlowInstanceId)
         val currentStep = formFlowInstance.getCurrentStep()
-
-        val expressionProcessor = expressionProcessorFactory.create(currentStep)
+        val expressionProcessor = expressionProcessorFactory.create(mapOf("step" to currentStep))
```
marijnritense, line 53: Having a bit of trouble getting my point across, but I don't think we should directly add the current step to the context. There's a lot of information in there that is irrelevant to the context, and as a result I think is better off not in there altogether. Maybe instead a DTO that contains the relevant data to that particular step? So something like the submission data for each form flow step instance up until the current step?

form-flow/src/main/kotlin/com/ritense/formflow/service/FormFlowService.kt
```diff
@@ -39,13 +40,20 @@ class FormFlowService(
     }
 
     fun open(formFlowInstanceId: FormFlowInstanceId) {
+        processExpressions(formFlowInstanceId, FormFlowStep::onOpen)
+    }
+
+    fun complete(formFlowInstanceId: FormFlowInstanceId) {
```
marijnritense, line 46: Shouldn't this get triggered when a step is completed? Now, in order to complete a step, you need to first make a call to FormFlowInstance.complete, and afterwards make a call to FormFlowService.complete... Either a call needs to be made from this complete method to the instance complete method, or we move this logic there after all.

theo-ritense: Yes, @Klaas-Ritense pointed out the same issue after submitting the PR. Still figuring out the right solution.

form-flow/src/main/kotlin/com/ritense/formflow/service/FormFlowService.kt
theo-ritense: Agreed. I can add an 'empty' DTO there of which the contents should be refined in another story.

marijnritense: Sounds good to me!

theo-ritense: I ended up just adding a basic variable map. Adding a whole conversion-layer that can be extended by implementation teams is way out of scope for this story (although desirable I guess)

form-flow/src/main/kotlin/com/ritense/formflow/service/FormFlowService.kt
theo-ritense: done

form-flow/src/main/kotlin/com/ritense/formflow/domain/instance/FormFlowStepInstance.kt
```diff
@@ -45,6 +47,40 @@ data class FormFlowStepInstance(
...
+    }
+
+    fun complete(submissionData: String) {
+        this.submissionData = submissionData
+
+        processExpressions(FormFlowStep::onComplete)
+    }
+
+    private fun processExpressions(expressionist: (FormFlowStep)-> List<String>?) {
+        ExpressionProcessorFactoryHolder.getinstance()?.let {
```
Klaas-Ritense, line 61: The ExpressionProcessorFactoryHolder is one of many ways to call a bean from a domain object. 

In my opinion, you should never call a bean from a domain object. But it seems no one agree with me. So if we really want to do that, we should at least sit together and decide on a single way to call beans from a domain class. Because I've seen several ways now

marijnritense (approved):

form-flow/src/main/kotlin/com/ritense/formflow/service/FormFlowService.kt
marijnritense: Makes sense. Extending this is out of scope for this feature for now anyway.

form-flow/src/main/kotlin/com/ritense/formflow/domain/instance/FormFlowStepInstance.kt
Klaas-Ritense, line 61: The ExpressionProcessorFactoryHolder is one of many ways to call a bean from a domain object. 

In my opinion, you should never call a bean from a domain object. But it seems no one agree with me. So if we really want to do that, we should at least sit together and decide on a single way to call beans from a domain class. Because I've seen several ways now

Klaas-Ritense (changes requested):

theo-ritense:
@Klaas-Ritense 
Sorry, I'm not able to reply directly on your [comment](https://github.com/valtimo-platform/valtimo-backend-libraries/pull/165#discussion_r850164608). Thanks GitHub: really convenient.

For me, it was about choosing the lesser evil. Having a complete() method in de domain entities (flow instance + step instance) AND in the service just calls for inconsistencies in future developments.

I'm interested in other solutions to call beans from the domain model. I've seen a few, but most of these rely on Spring and/or AOP magic which is hard to test.

Klaas-Ritense (approved):

form-flow/src/main/kotlin/com/ritense/formflow/domain/instance/FormFlowStepInstance.kt
Klaas-Ritense: Has been discussed

PR #168, opened 2022-04-14
added nextStep.open() to demo endpoint

theo-ritense:
- added nextStep.open() to demo endpoint
- removed unused method

### DIFF
--- a/form-flow-valtimo/src/main/kotlin/com/ritense/valtimo/formflow/web/rest/FormFlowResource.kt
+++ b/form-flow-valtimo/src/main/kotlin/com/ritense/valtimo/formflow/web/rest/FormFlowResource.kt
@@ -31,6 +31,7 @@ import org.springframework.web.bind.annotation.RequestMapping
 import org.springframework.web.bind.annotation.RestController
 import java.util.UUID
 import javax.transaction.Transactional
+import org.springframework.web.bind.annotation.RequestParam
 
 @RestController
 @RequestMapping(value = ["/api/form-flow/demo"])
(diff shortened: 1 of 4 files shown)


marijnritense (commented):
Can you also add it to createInstance?

Rick-Ritense (approved):

Rick-Ritense (approved):

form-flow-valtimo/src/main/kotlin/com/ritense/valtimo/formflow/web/rest/FormFlowResource.kt
```diff
@@ -60,6 +61,7 @@ class FormFlowResource(
     fun completeStep(
         @PathVariable(name = "instanceId") instanceId: String,
         @PathVariable(name = "stepId") stepId: String,
+        @RequestParam(name = "openNext", defaultValue = "false") openNext: Boolean = false,
```
Klaas-Ritense, line 64: How are we going to handle this scenario:
1. User fills in form
2. User completes the form
3. ???
4. User closes window
5. 10 min later, the user opens window again 
6. ??? 

The 'onOpen' expressions, are they executed at step 3 or step 6? or both?

ivo-ritense: Is that a scenario we have to cover though? Especially for the demo endpoints?

Klaas-Ritense: Im justwondering if we're all on the same page here. This solution seems to suggest that it behaves as an 'onCreate' rather that an 'onOpen'. Because I think it should happen at both step 3 and step 6

marijnritense: For the demo it happens in the same REST call (at least that's how it's written right now). I think for next sprint we want to look at the necessary APIs required to do all actions (completing a step, opening it, creating an instance(?)), so then we can just make the split properly.

PR #176, opened 2022-05-03
added form-type properties to step

theo-ritense:


Rick-Ritense (approved):

marijnritense:
I think for demo purposes it makes sense to update the existing endpoints to also return the form if present.

PR #178, opened 2022-05-06
34952 fixed equals inconsistencies

theo-ritense:
- added type field to equals comparison
- removed singular nextStep from configuration
- made some columns non-nullable to keep code cleaner

### DIFF
--- a/form-flow/src/main/resources/config/form-flow/schema/formflow.schema.json
+++ b/form-flow/src/main/resources/config/form-flow/schema/formflow.schema.json
@@ -75,10 +75,6 @@
                                 "step"
                             ]
                         }
-                    },
-                    "nextStep:": {
-                        "type": "string",
-                        "description": "The key of the next step in the flow. When multiple steps should be available you should use the property 'nextSteps'"
                     }
                 },
                 "required": [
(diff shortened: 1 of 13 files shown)


form-flow/src/main/resources/config/form-flow/schema/formflow.schema.json
```diff
@@ -75,10 +75,6 @@
                                 "step"
                             ]
                         }
-                    },
-                    "nextStep:": {
-                        "type": "string",
-                        "description": "The key of the next step in the flow. When multiple steps should be available you should use the property 'nextSteps'"
                     }
                 },
                 "required": [
```
Rick-Ritense, line 80: You can now make nextSteps required right?

theo-ritense: ah, yes. Will do

theo-ritense: I did say yes at first, but a final step will not have next steps.

Rick-Ritense: ah yeah of course

Rick-Ritense (approved):

PR #179, opened 2022-05-09
35029 - Created endpoint to retrieve form flow state

Rick-Ritense:
By sending an form flow instance id the endpoint will return the current step of that instance and the step definition type

### DIFF
--- a/form-flow-valtimo/src/test/kotlin/com/ritense/valtimo/formflow/web/rest/FormFlowResourceTest.kt
+++ b/form-flow-valtimo/src/test/kotlin/com/ritense/valtimo/formflow/web/rest/FormFlowResourceTest.kt
@@ -1,0 +1,91 @@
+package com.ritense.valtimo.formflow.web.rest
+
+import com.nhaarman.mockitokotlin2.mock
+import com.nhaarman.mockitokotlin2.whenever
+import com.ritense.formflow.domain.definition.FormFlowDefinition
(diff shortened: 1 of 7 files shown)


form-flow-valtimo/src/test/kotlin/com/ritense/valtimo/formflow/web/rest/FormFlowResourceTest.kt
```diff
@@ -0,0 +1,91 @@
...
+        formFlowInstance = mock()
+        whenever(formFlowInstance.id).thenReturn(formFlowInstanceId)
+        whenever(formFlowService.getInstanceById(formFlowInstance.id)).thenReturn(formFlowInstance)
+        formFlowResource = FormFlowResource(formFlowService)
+
+        mockMvc = MockMvcBuilders.standaloneSetup(formFlowResource).build()
+    }
+
+    @Test
+    fun `should return form flow state`() {
```
marijnritense, line 47: Does this provide enough coverage? Right now all that's tested is what happens when everything goes right, but I think testing failure states might be good too.

Rick-Ritense: Will add more test(s)

Klaas-Ritense (approved):

PR #180, opened 2022-05-09
#35016 prefill form in form flow with previously submitted data

Klaas-Ritense:


form/src/main/java/com/ritense/form/service/impl/FormIoFormLoaderService.java
```diff
@@ -56,4 +56,13 @@ public Optional<JsonNode> getFormDefinitionByNamePreFilled(final String formDefi
             );
     }
 
-}
\ No newline at end of file
+    @Override
+    public Optional<JsonNode> getFormDefinitionByNamePreFilled(final String formDefinitionName, final JsonNode content) {
```
marijnritense, line 60: I don't see this making use of the document for prefilling. Any way we can incorporate that here?

form-flow/src/main/kotlin/com/ritense/formflow/autoconfigure/FormFlowAutoConfiguration.kt
```diff
@@ -79,9 +80,10 @@ class FormFlowAutoConfiguration {
     fun formFlowService(
         formFlowDefinitionRepository: FormFlowDefinitionRepository,
         formFlowInstanceRepository: FormFlowInstanceRepository,
-        expressionProcessorFactory: ExpressionProcessorFactory
+        expressionProcessorFactory: ExpressionProcessorFactory,
```
Rick-Ritense, line 83: The expressionProcessorFactory doesn't seem to be used in creating the instance of FormFlowService so it can be removed

form/src/main/java/com/ritense/form/service/impl/FormIoFormLoaderService.java
Klaas-Ritense: Thats outside the scope of the story. But I added it anyway

form-flow-valtimo/src/main/kotlin/com/ritense/valtimo/formflow/web/rest/FormFlowResource.kt
```diff
@@ -46,8 +46,8 @@ class FormFlowResource(
                 FormFlowStepResult(
                     stepInstance.id.id,
                     stepInstance.definition.type.name,
-                    stepInstance.definition.type.properties
-                    // TODO Include the prefilled form when this functionality is available
+                    formFlowService.getFormFlowStepTypeHandler(stepInstance.definition.type)
```
marijnritense, line 49: Can also do this in 1 call, so you just say `getTypeProperties`, which will determine which handler should be used. Just not sure which one is "better"... Do you have any thoughts on this?

form-flow-valtimo/src/main/kotlin/com/ritense/valtimo/formflow/handler/FormFlowStepTypeFormHandler.kt
```diff
@@ -0,0 +1,82 @@
...
+            .orElseThrow { IllegalStateException("No FormDefinition found by name $formDefinitionName") }
+    }
+
+    private fun prefillWithSubmissionData(formDefinition: FormDefinition, stepInstance: FormFlowStepInstance) {
+        formDefinition.preFill(objectMapper.get().readTree(stepInstance.instance.getSubmissionDataContext()))
+    }
+
+    private fun prefillWithAdditionalData(
+        formDefinition: FormIoFormDefinition,
+        additionalParameters: Map<String, Any>
```
ivo-ritense, line 63: Was it not possible to use the existing prefill method to do this part part of the prefilling? That will prevent inconsistency issues if that function ever changes.

form-flow-valtimo/src/main/kotlin/com/ritense/valtimo/formflow/handler/FormFlowStepTypeFormHandler.kt
```diff
@@ -0,0 +1,82 @@
...
+    ) {
+        val documentId = additionalParameters["documentId"] as String?
+        val taskInstanceId = additionalParameters["taskInstanceId"] as String?
+
+        if (documentId == null) {
```
ivo-ritense, line 68: Do we want to not prefill the rest (i.e. taskId) if a documentId is not available?

ivo-ritense (approved):

form-flow-valtimo/src/main/kotlin/com/ritense/valtimo/formflow/handler/FormFlowStepTypeFormHandler.kt
Klaas-Ritense: That is only possible when fundamentally changing the existing prefill methods. And I prefer to not do that.
Right now they all need the document.

form-flow-valtimo/src/main/kotlin/com/ritense/valtimo/formflow/handler/FormFlowStepTypeFormHandler.kt
```diff
@@ -0,0 +1,82 @@
...
+
+    private fun prefillWithSubmissionData(formDefinition: FormDefinition, stepInstance: FormFlowStepInstance) {
+        formDefinition.preFill(objectMapper.get().readTree(stepInstance.instance.getSubmissionDataContext()))
+    }
+
+    private fun prefillWithAdditionalData(
+        formDefinition: FormIoFormDefinition,
+        additionalParameters: Map<String, Any>
+    ) {
+        val documentId = additionalParameters["documentId"] as String?
```
marijnritense, line 65: Almost forgot about this. documentId is not stored in the additionalParameters as such. Instead we just save the businesskey as `processInstanceBusinessKey`. There's 2 things that can be done here I think. Either upon creating the FormFlowInstance, we check if a document with this ID can be found already, and then save it in the additionalProperties as well. Alternatively, you can make the check yourself in this prefill. My preference goes to the first option, what do you think?

PR #187, opened 2022-05-13
#35298 form flow back

Klaas-Ritense:


form-flow-valtimo/src/main/kotlin/com/ritense/valtimo/formflow/web/rest/FormFlowResource.kt
```diff
@@ -91,4 +91,27 @@ class FormFlowResource(
...
+    @PostMapping("/{formFlowId}/back", produces = [MediaType.APPLICATION_JSON_VALUE])
+    @Transactional
+    fun backStep(
+        @PathVariable(name = "formFlowId") formFlowId: String
+    ): ResponseEntity<CompleteStepResult> {
+        val instance = formFlowService.getByInstanceIdIfExists(
+            FormFlowInstanceId.existingId(UUID.fromString(formFlowId))
+        )!!
+
+        val stepInstance = instance.back()
```
ivo-ritense, line 104: is the form flow instance saved?

form-flow/src/main/kotlin/com/ritense/formflow/domain/instance/FormFlowInstance.kt
```diff
@@ -142,6 +150,7 @@ class FormFlowInstance(
             this.currentFormFlowStepInstanceId = null
             return null
         }
+        history.removeAll { it.order >= nextStep.order }
```
ivo-ritense, line 153: What we discussed about this in the past is that we want to keep the history as long as it's relevant. this means that we want to only remove history for future steps after going back when another "path" is chosen (i.e. because of conditions another nextstep is activated. This prevents data loss when going back and then to next step again.

form-flow/src/main/kotlin/com/ritense/formflow/domain/instance/FormFlowInstance.kt
```diff
@@ -152,19 +161,19 @@ class FormFlowInstance(
             return null
         }
         var stepKey = formFlowDefinition.startStep
+        var stepOder = 0
```
ivo-ritense, line 164: stepOrder spelling mistake

form-flow/src/main/kotlin/com/ritense/formflow/domain/instance/FormFlowInstance.kt
```diff
@@ -152,19 +161,19 @@ class FormFlowInstance(
...
-            if (currentStep.nextSteps == null || currentStep.nextSteps.isEmpty()) {
+            if (currentStep.nextSteps.isEmpty()) {
                 return null
             }
 
             stepKey = currentStep.nextSteps.first().step
+            stepOder = currentStepInstance.order + 1
         }
-        return FormFlowStepInstance(instance = this, stepKey = stepKey, order = history.size)
+        return FormFlowStepInstance(instance = this, stepKey = stepKey, order = stepOder)
```
ivo-ritense, line 176: We should return the existing step if one exists

ivo-ritense (approved):

PR #190, opened 2022-05-13
Set up form flow test config for gzac app

ivo-ritense:


app/gzac/src/main/resources/config/form-flow/test-form-flow.json
```diff
@@ -0,0 +1,65 @@
+{
```
Klaas-Ritense, line 1: Gzac? not in valtimo-core?

Klaas-Ritense (approved):

app/gzac/src/main/resources/config/form-flow/test-form-flow.json
ivo-ritense: That was easier for my test case. You can also include it in the valtimo app if you want.

PR #195, opened 2022-05-18
#35377 [Ticket 34650] Document aangemaakt door: 'automatisch gegenereerd'

Klaas-Ritense:


document/src/main/java/com/ritense/document/domain/impl/listener/DocumentRelatedFileSubmittedEventListenerImpl.java
```diff
@@ -38,7 +39,7 @@ public void handle(DocumentRelatedFileSubmittedEvent event) {
         var resource = resourceService.getResource(event.getResourceId());
         documentService.assignRelatedFile(
             JsonSchemaDocumentId.existingId(event.getDocumentId()),
-            JsonSchemaRelatedFile.from(resource)
+            JsonSchemaRelatedFile.from(resource).withCreatedBy(SecurityUtils.getCurrentUserLogin())
```
marijnritense, line 42: What will this return if this is indeed generated automatically? I think those should still be named the same, right?

marijnritense: Also, I'd like to see some tests.

Klaas-Ritense: `SecurityUtils.getCurrentUserLogin()` returns 'null' if it was automatically generated. 
The FE will replaces a 'null' with 'Automatisch gegenereerd'

Rick-Ritense (approved):

marijnritense (approved):

PR #197, opened 2022-05-20
#35343 Creating the validator can take up to 5 seconds on some environments.

Klaas-Ritense:


contract/src/main/kotlin/com/ritense/valtimo/contract/validation/autoconfigure/ValidatorAutoConfiguration.kt
```diff
@@ -0,0 +1,42 @@
...
+import org.springframework.context.annotation.Configuration
+import javax.validation.Validation
+import javax.validation.Validator
+
+@Configuration
+class ValidatorAutoConfiguration {
+
+    @Bean
+    @ConditionalOnMissingBean(Validator::class)
+    fun validator(): Validator {
```
marijnritense, line 32: I don't think this has to be a bean, can just initialize it in the companion class or the onApplicationReady method for the ValidatorReadyEventListener

contract/src/main/kotlin/com/ritense/valtimo/contract/validation/Validatable.kt
```diff
@@ -24,17 +24,26 @@ import javax.validation.Validator
...
         }
     }
 
-}
\ No newline at end of file
+    companion object {
+        private var validator: Validator? = null
+
+        fun getValidator(): Validator {
+            return validator ?: Validation.buildDefaultValidatorFactory().validator
```
theo-ritense, line 41: Why not set it when it's null? You left a possibility open where it would still create a new validator every request.

theo-ritense (changes requested):

contract/src/main/kotlin/com/ritense/valtimo/contract/validation/Validatable.kt
marijnritense: Fixed

theo-ritense (approved):

PR #209, opened 2022-06-17
Load plugins on startup

ivo-ritense:


plugin/src/main/kotlin/com/ritense/plugin/PluginDeploymentListener.kt
```diff
@@ -0,0 +1,48 @@
...
+
+class PluginDeploymentListener(
+    private val pluginDefinitionResolver: PluginDefinitionResolver,
+    private val pluginDefinitionRepository: PluginDefinitionRepository
+) {
+
+    @EventListener(ApplicationStartedEvent::class)
+    fun deployPluginDefinitions() {
+        val classes = findPluginClasses()
+        val action: (Map.Entry<Class<*>, Plugin>) -> Unit = { (clazz, pluginAnnotation) ->
```
marijnritense, line 33: I feel like "action" is a bit of a loaded term. maybe rename this variable (or just omit naming this variable altogether, I don't think this matters much for readability).

theo-ritense: Agreed, action is more a term used for the PluginAction (methods)

plugin/src/main/kotlin/com/ritense/plugin/PluginDeploymentListener.kt
```diff
@@ -0,0 +1,48 @@
...
+import org.springframework.boot.context.event.ApplicationStartedEvent
+import org.springframework.context.event.EventListener
+
+class PluginDeploymentListener(
+    private val pluginDefinitionResolver: PluginDefinitionResolver,
+    private val pluginDefinitionRepository: PluginDefinitionRepository
+) {
+
+    @EventListener(ApplicationStartedEvent::class)
+    fun deployPluginDefinitions() {
```
marijnritense, line 31: Would like to see some logging here, both to mention deployment is happening, and to state which plugins were (newly) deployed.

plugin/src/test/kotlin/com/ritense/plugin/PluginDeploymentListenerIT.kt
```diff
@@ -0,0 +1,41 @@
...
+import org.junit.jupiter.api.Test
+import org.springframework.beans.factory.annotation.Autowired
+
+internal class PluginDeploymentListenerIT: BaseIntegrationTest() {
+
+    @Autowired
+    lateinit var pluginDefinitionRepository: PluginDefinitionRepository
+
+    @Test
+    fun `should deploy test plugin`() {
```
marijnritense, line 30: What happens if a plugin cannot be deployed? I think we need to have that be covered by a test as well. What kind of output are we expecting? What should be logged? Should the application stop (fail-fast), or just keep starting up?

plugin/src/main/kotlin/com/ritense/plugin/PluginDeploymentListener.kt
ivo-ritense: fixed

plugin/src/test/kotlin/com/ritense/plugin/PluginDeploymentListenerIT.kt
ivo-ritense: fixed

plugin/src/main/kotlin/com/ritense/plugin/PluginDeploymentListener.kt
ivo-ritense: fixed

plugin/src/main/kotlin/com/ritense/plugin/annotation/Plugin.kt
```diff
@@ -0,0 +1,25 @@
...
+ */
+
+package com.ritense.plugin.annotation
+
+@Target(AnnotationTarget.CLASS)
+@Retention(AnnotationRetention.RUNTIME)
+annotation class Plugin(
+    val key: String,
+    val title: String,
+    val description: String,
```
Rick-Ritense, line 24: Should the title and the description not be arranged on the frontend, because of translations?

marijnritense: This can still be done by key. All this does is provide a default.

Rick-Ritense: Is the backend responsible for that?

marijnritense: In my opinion, if FE does not have any translations available, then it makes sense for BE to provide a default.

marijnritense (approved):

PR #212, opened 2022-06-23
Feature/oa prefix

erik-ritense:
In a Valtimo implemetnation we want to read and write data from a form to the objects-api (https://github.com/maykinmedia/objects-api#objects-api) 

For this we like to use the oa prefix as ExternalFormFieldType. This PR makes the oa prefix available within Valtimo (without an implementation).


gradle.properties
```diff
@@ -26,4 +26,4 @@ org.gradle.parallel=true
 # override liquibase to fix bug see :
 # https://forum.liquibase.org/t/liquibaseexception-java-lang-classcastexception-class-java-time-localdatetime-cannot-be-cast-to-class-java-lang-string/5059/3
 liquibase.version=4.3.3
-projectVersion=9.13.0.RELEASE
+projectVersion=9.14.0.124.RC-SNAPSHOT
```
marijnritense, line 29: Don't think this needs to be set. Can this be reverted?

objects-api/src/main/kotlin/com/ritense/objectsapi/listener/ObjectsApiFormDataListener.kt
```diff
@@ -0,0 +1,13 @@
...
+
+import com.ritense.document.exception.NotImplementedException
+import com.ritense.valtimo.contract.event.ExternalDataSubmittedEvent
+import org.springframework.context.event.EventListener
+
+class ObjectsApiFormDataListener {
+
+    @EventListener(ExternalDataSubmittedEvent::class)
+    fun handle(event: ExternalDataSubmittedEvent) {
+        throw NotImplementedException()
```
marijnritense, line 11: This will throw exceptions every time an ExternalDataSubmittedEvent is published. I think this can be considered a breaking change. Maybe log something instead?

objects-api/src/main/kotlin/com/ritense/objectsapi/resolver/ObjectsApiFormDataResolver.kt
```diff
@@ -0,0 +1,17 @@
...
+import java.util.*
+
+class ObjectsApiFormDataResolver: FormFieldDataResolver {
+
+    override fun supports(externalFormFieldType: ExternalFormFieldType): Boolean {
+        return externalFormFieldType == ExternalFormFieldType.OA
+    }
+
+    override fun get(documentDefinitionName: String, documentId: UUID, vararg varNames: String): Map<String, Any> {
+        throw NotImplementedException()
```
marijnritense, line 15: This feels a bit weird. This implements FormFieldDataResolver, but then throws a NotImplementedException. Why is this?

objects-api/src/main/kotlin/com/ritense/objectsapi/resolver/ObjectsApiFormDataResolver.kt
```diff
@@ -0,0 +1,17 @@
+package com.ritense.objectsapi.resolver
+
+import com.ritense.document.exception.NotImplementedException
+import com.ritense.valtimo.contract.form.ExternalFormFieldType
+import com.ritense.valtimo.contract.form.FormFieldDataResolver
+import java.util.*
```
marijnritense, line 6: Checkstyle

marijnritense (approved):

PR #220, opened 2022-06-28
Bugfix/too many form link results are ignored

Tom-Ritense:
Introduced single function to ensure retrieval is always returning 1 and only 1 otherwise an error state is masked.

Solution distilled from https://stackoverflow.com/questions/22694884/filter-java-stream-to-1-and-only-1-element


form-link/src/main/java/com/ritense/formlink/service/impl/CamundaFormAssociationService.java
```diff
@@ -453,4 +458,17 @@ private ObjectNode buildMetaDataObject(Document document) {
         metaDataNode.put("sequence", document.sequence());
         return metaDataNode;
     }
+
+    private static <T> Collector<T, ?, Optional<T>> single() {
```
marijnritense, line 462: Not entirely happy with this name, as it doesn't fully convey what this method is supposed to return. It returns a collector that results in one result if the list only has one entry. That's not clear when reading the method name.

Tom-Ritense: Do you have a suggestion for improving the name? from the stackoverflow I see a couple:

- onlyElement see [
](https://guava.dev/releases/snapshot/api/docs/com/google/common/collect/MoreCollectors.html#onlyElement--)- singleElementCollector
- zeroOrOne
- onlyOne
- toSingleton

marijnritense: singleElementCollector sounds best out of these options. I was playing with names like singleElementIfOnlyOneElementPresentCollector, but that becomes too cumbersome in my opinion.

form-link/src/main/java/com/ritense/formlink/service/impl/CamundaFormAssociationService.java
```diff
@@ -453,4 +458,17 @@ private ObjectNode buildMetaDataObject(Document document) {
...
+        return Collectors.collectingAndThen(
+            Collectors.toList(),
+            list -> {
+                if (list.size() == 1) {
+                    return Optional.of(list.get(0));
+                }
+                throw new IllegalStateException("Expected single result found: " + list.size());
```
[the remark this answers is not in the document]
Tom-Ritense: Ok will change.

form-link/src/main/java/com/ritense/formlink/service/impl/CamundaFormAssociationService.java
Tom-Ritense: singleElementCollector is fine then.

marijnritense (approved):

PR #221, opened 2022-06-29
#35792 Implementatie SmartDocuments plugin

Klaas-Ritense:


ivo-ritense (approved):

document-generation/smartdocuments/src/main/kotlin/com/ritense/smartdocuments/plugin/SmartDocumentsPluginFactory.kt
```diff
@@ -0,0 +1,47 @@
...
+    private val pluginService: PluginService,
+    private val documentService: DocumentService,
+    private val resourceService: ResourceService,
+    private val processDocumentService: ProcessDocumentService,
+    private val applicationEventPublisher: ApplicationEventPublisher,
+    private val smartDocumentsClient: SmartDocumentsClient,
+) : PluginFactory<SmartDocumentsPlugin> {
+
+    override fun create(pluginConfigurationKey: String): SmartDocumentsPlugin {
+        val configuration = pluginService.getPluginConfiguration(pluginConfigurationKey)
```
ivo-ritense, line 37: I think it would be better to always pass in an instance of the PluginConfiguration instead of the key so every factory doesn't have to implement this line.

Klaas-Ritense: Changed it

marijnritense (approved):

PR #222, opened 2022-06-29
bugfix form association error: 'Expected single result but found: 0'

Klaas-Ritense:


form-link/src/main/java/com/ritense/formlink/service/impl/CamundaFormAssociationService.java
```diff
@@ -463,7 +463,9 @@ private ObjectNode buildMetaDataObject(Document document) {
         return Collectors.collectingAndThen(
             Collectors.toList(),
             list -> {
-                if (list.size() == 1) {
+                if (list.isEmpty()) {
+                    return Optional.empty();
```
Tom-Ritense, line 467: Dus als een lijst al leeg is dan gaan we ervan uit dat deze functie succesvol is afgerond? Ik twijfel of dit niet een error state is.

Tom-Ritense: I can the list is initially empty the singleElementCollector will fail which is mitigated with the check if the list is empty.

Tom-Ritense (approved):

form-link/src/main/java/com/ritense/formlink/service/impl/CamundaFormAssociationService.java
Tom-Ritense: Maybe add this test case:
``
   @Test
    public void shouldGetEmptyFormAssociationByFormLinkIdWhenZeroExist() {
        final var formAssociations = new FormAssociations();
        var camundaProcessFormAssociation = new CamundaProcessFormAssociation(
            CamundaProcessFormAssociationId.newId(UUID.randomUUID()),
            PROCESS_DEFINITION_KEY,
            formAssociations
        );
        when(processFormAssociationRepository.findByProcessDefinitionKey(eq(PROCESS_DEFINITION_KEY)))
            .thenReturn(Optional.of(camundaProcessFormAssociation));

        var value = camundaFormAssociationService.getFormAssociationByFormLinkId(PROCESS_DEFINITION_KEY, "user-task-id")
        assertThat(value.isEmpty()).isTrue();
    }

PR #233, opened 2022-07-07
Feature/35829 initializing plugins

marijnritense:


plugin/src/main/kotlin/com/ritense/plugin/PluginDeploymentListener.kt
```diff
@@ -41,40 +45,87 @@ class PluginDeploymentListener(
...
+        pluginProperties: MutableSet<PluginProperty>,
+        clazz: Class<*>
+    ) {
+        val properties = findPluginProperties(clazz)
+        properties.forEach { (field, propertyAnnotation) ->
+            pluginProperties.add(
+                PluginProperty(
+                    PluginPropertyId(
+                        propertyAnnotation.key,
+                        pluginDefinition
```
ivo-ritense, line 84: I'm wondering if we want to control this relation from one side. i.e. have the constructor call an pluginDefinition.addPluginProperty(this) method or something

plugin/src/main/kotlin/com/ritense/plugin/PluginDeploymentListener.kt
```diff
@@ -41,40 +45,87 @@ class PluginDeploymentListener(
...
+        }
+    }
+
     private fun findPluginClasses() : Map<Class<*>, Plugin> {
         return pluginDefinitionResolver.findPluginClasses()
     }
 
     private fun deployPluginDefinition(pluginDefinition: PluginDefinition): PluginDefinition {
         logger.info { "Deploying plugin ${pluginDefinition.key}" }
+        logger.debug { "$pluginDefinition" }
```
ivo-ritense, line 119: What does this output look like? is it a proper toString, or does it only output the type+hash like this?

document-generation/smartdocuments/src/main/kotlin/com/ritense/smartdocuments/plugin/SmartDocumentsPlugin.kt
```diff
@@ -58,9 +59,11 @@ class SmartDocumentsPlugin(
     private val processDocumentService: ProcessDocumentService,
     private val applicationEventPublisher: ApplicationEventPublisher,
     private val smartDocumentsClient: SmartDocumentsClient,
-    private val smartDocumentsPluginProperties: SmartDocumentsPluginProperties,
 ) {
 
+    @PluginProperty(key = "smartDocumentsPluginProperties")
+    private val smartDocumentsPluginProperties: SmartDocumentsPluginProperties? = null
```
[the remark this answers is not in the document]
ivo-ritense: i feel like it might be. it means you don't have to deal with the null check every time you use the variable.

plugin/src/test/kotlin/com/ritense/plugin/PluginFactoryTest.kt
```diff
@@ -0,0 +1,134 @@
...
+            "title",
+            ObjectMapper().valueToTree("{\"name\": \"whatever\" }"),
+            pluginDefinition
+        )
+
+        assertTrue(pluginFactory.canCreate(pluginConfiguration))
+    }
+
+    @Test
+    fun `should not be able to create plugin instance`() {
```
ivo-ritense, line 58: i would add the condition in which it shouldn't work to the method name. e.g. `should not be able to create plugin instance for different plugin definition`

ivo-ritense (approved):

PR #243, opened 2022-07-13
Postgres DB doesn't support 'unhex' and doesn't support "" for strings

Klaas-Ritense:


form-link/src/main/resources/config/liquibase/changelog/20220630-storage-performance-improvement-changelog.xml
```diff
@@ -81,23 +81,40 @@
...
 
+    <changeSet author="Ritense" id="3-1-2" dbms="postgresql">
+        <sql>
+            INSERT INTO process_form_association (
+                process_form_association_id,
+                process_definition_key,
+                form_associations
+            )
+            VALUES (
+                DECODE(REPLACE('77dea2ad-3c8c-40c6-a278-7cf1a1ac9384', '-', ''), 'hex'),
```
ivo-ritense, line 92: hardcoded uuid?

Klaas-Ritense: Copied from the cangeset '3-1' above. 
This hardcoded uuid is deleted at the changeset '3-3' below.

According to Tom, there are meant to test the '3-2' changeset

PR #252, opened 2022-07-20
Bugfix Plugin deployment

Klaas-Ritense:
- Plugins were not always deployed. This broke the IntegrationTests. Cause: TransactionalEventListener only runs when inside a Transaction.

- Stackoverflow due to ManyToOne relation. Fix: first remove all plugin-properties/action/action-properties. Because these have a ManyToOne relation with ProcessDefinition.

- Occasional error in IntelliJ: `Expiring Daemon because JVM heap space is exhausted`. FIx: additional jvmargs.

### DIFF
--- a/plugin/src/main/kotlin/com/ritense/plugin/PluginDeploymentListener.kt
+++ b/plugin/src/main/kotlin/com/ritense/plugin/PluginDeploymentListener.kt
@@ -16,7 +16,6 @@
 
 package com.ritense.plugin
 
-import com.ritense.plugin.annotation.PluginProperty as PluginPropertyAnnotation
 import com.ritense.plugin.annotation.Plugin
 import com.ritense.plugin.annotation.PluginAction
 import com.ritense.plugin.annotation.PluginActionProperty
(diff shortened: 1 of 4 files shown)


plugin/src/main/kotlin/com/ritense/plugin/PluginDeploymentListener.kt
```diff
@@ -29,23 +28,30 @@ import com.ritense.plugin.exception.PluginDefinitionNotDeployedException
...
+    private val pluginPropertyRepository: PluginPropertyRepository,
     private val pluginActionDefinitionRepository: PluginActionDefinitionRepository,
     private val pluginActionPropertyDefinitionRepository: PluginActionPropertyDefinitionRepository
 ) {
 
-    @TransactionalEventListener(ApplicationStartedEvent::class)
+    @EventListener(ApplicationStartedEvent::class)
     fun deployPluginDefinitions() {
         logger.info { "Deploying plugins" }
+        pluginPropertyRepository.deleteAll()
```
marijnritense, line 51: This causes a lot of unnecessary deletions, maybe change it so deletions only happen if there were changes?

plugin/src/main/kotlin/com/ritense/plugin/PluginDeploymentListener.kt
```diff
@@ -29,23 +28,30 @@ import com.ritense.plugin.exception.PluginDefinitionNotDeployedException
...
 class PluginDeploymentListener(
     private val pluginDefinitionResolver: PluginDefinitionResolver,
     private val pluginDefinitionRepository: PluginDefinitionRepository,
+    private val pluginPropertyRepository: PluginPropertyRepository,
     private val pluginActionDefinitionRepository: PluginActionDefinitionRepository,
     private val pluginActionPropertyDefinitionRepository: PluginActionPropertyDefinitionRepository
 ) {
 
-    @TransactionalEventListener(ApplicationStartedEvent::class)
+    @EventListener(ApplicationStartedEvent::class)
```
marijnritense, line 48: Wouldn't adding the transactional annotation to the test also work? That allows us to keep the annotation as-is here.

plugin/src/main/kotlin/com/ritense/plugin/PluginDeploymentListener.kt
Klaas-Ritense: unfortunately, that won't fix the stackoverflow error. This error happens even when nothing changed at all: On every second application start-up the and the plugins are deployed for the second time, the error happens

plugin/src/main/kotlin/com/ritense/plugin/PluginDeploymentListener.kt
Klaas-Ritense: After starting the application locally, I found that plugins were not deployed at all anymore.

marijnritense (approved):

PR #256, opened 2022-07-25
36041 Resolving of plugin configuration placeholder variables

theo-ritense:


plugin/src/test/kotlin/com/ritense/plugin/service/PluginServiceIT.kt
```diff
@@ -66,7 +78,38 @@ internal class PluginServiceIT: BaseIntegrationTest() {
...
+
+        val execution = DelegateExecutionFake.of()
+            .withProcessInstanceId(UUID.randomUUID().toString())
+
+        pluginService.invoke(execution, processLink)
+    }
+
+    @Test
+    @Transactional
+    fun `should invoke an action with a placeholder value on the plugin`() {
```
marijnritense, line 90: So this is the happy path, what if there is no resolver? This should also be covered by a test.

value-resolver/src/test/kotlin/com/ritense/valueresolver/ValueResolverFactoryServiceTest.kt
```diff
@@ -79,7 +78,27 @@ internal class ValueResolverFactoryServiceTest {
...
+                )
+            )
+        }
+
+        assertThat(exception.message).startsWith("No resolver factory found for value prefix xyz")
+    }
+
+    @Test
+    @Throws(RuntimeException::class)
+    fun `Should throw exception when no resolvers are configured`() {
```
theo-ritense, line 97: @marijnritense I added this to test the behaviour when no resolvers are found, as I don't think the plugin service is responsible for this.

plugin/src/test/kotlin/com/ritense/plugin/service/PluginServiceIT.kt
theo-ritense: please see https://github.com/valtimo-platform/valtimo-backend-libraries/pull/256/files#r928835211

marijnritense (approved):

PR #258, opened 2022-07-25
Encrypt plugin properties that are marked with encrypt in annotation

ivo-ritense:


plugin/src/main/kotlin/com/ritense/plugin/domain/PluginConfiguration.kt
```diff
@@ -39,9 +43,9 @@ class PluginConfiguration(
     var title: String,
     @Type(type = "com.vladmihalcea.hibernate.type.json.JsonType")
     @Column(name = "properties", columnDefinition = "JSON")
-    var properties: JsonNode? = null,
+    val properties: JsonNode? = null,
     @JoinColumn(name = "plugin_definition_key", updatable = false, nullable = false)
-    @ManyToOne(fetch = FetchType.LAZY)
+    @ManyToOne(fetch = FetchType.EAGER)
```
marijnritense, line 48: Why was this necessary? Do we always need the definition?

plugin/src/main/kotlin/com/ritense/plugin/domain/PluginConfiguration.kt
```diff
@@ -51,4 +55,20 @@ class PluginConfiguration(
             Mapper.INSTANCE.get().treeToValue(properties, T::class.java)
         }
     }
+
+    fun updateProperties(propertiesForUpdate: JsonNode) {
+        if (properties is ObjectNode) {
```
marijnritense, line 60: If it's not an `ObjectNode`, then it can't be updated. I would expect this method to throw an exception in that case.

plugin/src/main/kotlin/com/ritense/plugin/service/EncryptionService.kt
```diff
@@ -0,0 +1,42 @@
...
+import javax.crypto.spec.SecretKeySpec
+
+class EncryptionService(
+    secret: String
+) {
+    private var key = SecretKeySpec(secret.toByteArray(), AES)
+    private var cipher = Cipher.getInstance(AES)
+
+    fun encrypt(value: String): String {
+        cipher.init(Cipher.ENCRYPT_MODE, key)
```
marijnritense, line 30: This feels a bit dirty to me. If we always init it with the same arguments, then shouldn't we just have 2 ciphers? 1 for encrypting, 1 for decrypting.

plugin/src/main/kotlin/com/ritense/plugin/domain/PluginConfiguration.kt
ivo-ritense: Every time we load a plugin configuration we need the definition to identify any secrets we need to decrypt. Since we access this variable every time there's no reason to keep this lazy.

plugin/src/main/kotlin/com/ritense/plugin/domain/PluginConfiguration.kt
ivo-ritense: Since we need the properties to always be an objectnode, i changed the type to reflect that

plugin/src/main/kotlin/com/ritense/plugin/service/EncryptionService.kt
ivo-ritense: Fixed

marijnritense (approved):

PR #259, opened 2022-07-26
#36275 Temporary file storage

Klaas-Ritense:


resource/src/main/kotlin/com/ritense/resource/service/ResourceService.kt
```diff
@@ -23,11 +23,14 @@ import com.ritense.resource.web.ResourceDTO
 import com.ritense.valtimo.contract.resource.FileStatus
 import com.ritense.valtimo.contract.resource.Resource
 import org.springframework.web.multipart.MultipartFile
+import java.io.InputStream
 import java.net.URL
 import java.util.UUID
 
 interface ResourceService {
```
Klaas-Ritense, line 30: New and more generic methods for the ResourceService. Do you guys agree?

resource/temporary-resource/gradle/publishing.gradle
```diff
@@ -0,0 +1,35 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+pluginManager.withPlugin('maven-publish') {
+    publishing {
+        publications {
+            maven(MavenPublication) {
+                pom {
+                    name = 'Temporary resource module'
```
marijnritense, line 22: Does `temporary resource` make sense as a module name? I feel like `temporary local storage` might fit better, or `temporary file storage`, or `temporary resource storage`. Just something with storage in the name.

ivo-ritense (approved):

resource/temporary-resource/gradle/publishing.gradle
Klaas-Ritense: renamed

PR #266, opened 2022-07-29
fix: 36319 multiple forms linked to single start event

Klaas-Ritense:


form-link/src/main/resources/config/liquibase/changelog/20220630-storage-performance-improvement-changelog.xml
```diff
@@ -118,4 +118,16 @@
         </sql>
     </changeSet>
 
+    <changeSet author="Ritense" id="4">
+        <sql>
+            DELETE FROM process_form_association_v2
+            WHERE form_association_type = 'start-event' and form_association_id != (select * from(
```
ivo-ritense, line 124: is the extra `select * from` needed?

Klaas-Ritense: Its needed in mysql for some weird reason. See:
https://www.codeproject.com/Tips/831164/MySQL-can-t-specify-target-table-for-update-in-FRO

ivo-ritense: This filter doesn't work properly. it will only keep 1 form association, instead of 1 per process definition.

Klaas-Ritense: I tested this query locally and it works.
They subquery makes sure that only 1 start-event per process_definition is allowed

ivo-ritense (approved):

PR #269, opened 2022-08-02
Feature/support for custom external field resolver

Tom-Ritense:
Introducing support for creating ExternalFormFieldResolvers that implement a new support(String name) method.
The goal is to let implementation teams define their own resolvers outside the ExternalFormFieldType enum definition.
Form keys should be able to define a key like so "anyDataSource.propertyName". 
Note: the old support(ExternalFormFieldType) method is being deprecated.


form/src/main/java/com/ritense/form/domain/FormIoFormDefinition.java
```diff
@@ -320,42 +327,50 @@ private boolean isDocumentContentVar(JsonNode field) {
...
-        return false;
-    }
-
-    private Optional<ExternalFormFieldType> getExternalFormFieldType(JsonNode field) {
-        var propertyValue = field.get(PROPERTY_KEY).asText().toUpperCase();
-        for (ExternalFormFieldType externalFormFieldType : ExternalFormFieldType.values()) {
-            if (propertyValue.startsWith(externalFormFieldType.name().toUpperCase())) {
-                return Optional.of(externalFormFieldType);
+        // Check if key prefix is supported by a resolver.
+        final var resolvers = SpringContextHelper.getBeansOfType(FormFieldDataResolver.class);
```
marijnritense, line 368: Generally speaking this is a security risk, as now there's a static method that can retrieve any bean at every point in the code. I suggest making a package private variation of this `SpringContextHelper` that only retrieves `FormFieldDataResolver` beans.

document/src/main/java/com/ritense/document/config/SpringContextHelper.java
```diff
@@ -33,4 +35,9 @@ public void setApplicationContext(@NotNull ApplicationContext applicationContext
     public static <T> T getProperty(String property, Class<T> targetClass) {
         return context.getEnvironment().getProperty(property, targetClass);
     }
+
+    public static <T> Map<String, T> getBeansOfType(Class<T> targetClass) {
```
marijnritense, line 39: Generally speaking this is a security risk, as now there's a static method that can retrieve any bean at every point in the code. I suggest making a package private variation of this SpringContextHelper that only retrieves FormFieldDataResolver beans.

form/src/main/java/com/ritense/form/domain/FormIoFormDefinition.java
Tom-Ritense: Will do.

marijnritense (approved):

PR #270, opened 2022-08-05
Feature/zgw plugins

ivo-ritense:


zgw/documenten-api/src/main/kotlin/com/ritense/documentenapi/client/DocumentenApiClient.kt
```diff
@@ -0,0 +1,47 @@
...
+    val webclient: WebClient
+) {
+    fun storeDocument(
+        authentication: DocumentenApiAuthentication,
+        baseUrl: String,
+        request: CreateDocumentRequest
+    ): CreateDocumentResult {
+        val result = webclient
+            .mutate()
+            .filter(authentication)
```
marijnritense, line 35: Do we foresee usage of multiple filters for this client in the future? If so, then this solution isn't really future proof. Maybe instead we can accept a list of filters? This would make it so we don't have to worry about the issue altogether.

marijnritense: I mean in addition to accepting the `DocumentenApiAuthentication`, as that one is mandatory.

plugin/src/main/kotlin/com/ritense/plugin/service/PluginService.kt
```diff
@@ -215,7 +215,13 @@ class PluginService(
 
         return paramValues.mapValues { (param, value) ->
                 if (value != null && value.isTextual) {
-                    placeHolderValueMap.getOrDefault(value.textValue(), objectMapper.treeToValue(value, param.type))
+                    //TODO: possible issue here. what if value is textual but argument type isn't
```
marijnritense, line 218: Write a test that covers the situation?

openzaak/src/main/kotlin/com/ritense/openzaak/plugin/OpenZaakPlugin.kt
```diff
@@ -0,0 +1,51 @@
...
+)
+class OpenZaakPlugin(
+    val tokenGeneratorService: OpenZaakTokenGeneratorService
+): DocumentenApiAuthentication {
+    @PluginProperty(key = "clientId", secret = false, required = true)
+    lateinit var clientId: String
+    @PluginProperty(key = "clientSecret", secret = true, required = true)
+    lateinit var clientSecret: String
+
+    override fun filter(request: ClientRequest, next: ExchangeFunction): Mono<ClientResponse> {
```
marijnritense, line 41: Should probably write a test that checks this method actually sets the authentication. Fairly trivial, but should still be tested.

ivo-ritense: added test

marijnritense (approved):

PR #274, opened 2022-08-09
Added support for resource roles

ferdi-ritense:
Also retrieve resource roles from specified client. Story: https://ritense.tpondemand.com/entity/36616-be-add-possibility-to-get-roles


ivo-ritense (approved):
Made some changes, please see if you agree.

PR #281, opened 2022-08-18
Updated the Keycloak dependencies to resolve CVE-2022-1245

IvarKoreman-Ritense:
CVE-2022-1245 has been resolved in version 18.0.0. See https://nvd.nist.gov/vuln/detail/CVE-2022-1245

### DIFF
--- a/keycloak-iam/build.gradle
+++ b/keycloak-iam/build.gradle
@@ -45,8 +45,8 @@ dependencies {
 
     implementation group: 'org.springframework.boot', name: 'spring-boot-starter-web'
     implementation group: 'org.springframework.boot', name: 'spring-boot-starter-data-jpa'
-    implementation group: 'org.keycloak', name: 'keycloak-spring-boot-starter', version: '15.0.0'
-    implementation group: 'org.keycloak', name: 'keycloak-admin-client', version: '15.0.0'
+    implementation group: 'org.keycloak', name: 'keycloak-spring-boot-starter', version: '19.0.1'
+    implementation group: 'org.keycloak', name: 'keycloak-admin-client', version: '19.0.1'
 
     testImplementation project(":core")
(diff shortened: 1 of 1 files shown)


ivo-ritense (approved):

ivo-ritense:
Do you know if this is backwards compatible or not? Will this client work with older versions? Otherwise we might have to upgrade during the next major version.

IvarKoreman-Ritense:
> Do you know if this is backwards compatible or not? Will this client work with older versions? Otherwise we might have to upgrade during the next major version.

I have tested it on our own implementation and it seems to work fine. I don't see anything in the change logs of these dependencies that indicates that something cloud break, so it _seems_ like it will be backwards compatible, but I can't guarantee it.

The customer of this implementation wants this to be resolved before we take the application into production, but we can update this dependency this in our implementation, so we are currently not depending on a fix in Valtimo itself. Though I can imagine that we want this CVE resolved ASAP, so I'll leave the decision on whether this should be part of the major upgrade or not up to you guys. :)

The code that we are using to resolve this in our implementation:

```groovy
implementation "com.ritense.valtimo:keycloak-iam:${valtimo_version}"
constraints {
    implementation(group: 'org.keycloak', name: 'keycloak-spring-boot-starter', version: '19.0.1') {
        because 'CVE. This constraint can be removed when this dependency is upgraded in com.ritense.valtimo:keycloak-iam ' +
                'See  https://nvd.nist.gov/vuln/detail/CVE-2022-1245'
    }
    implementation(group: 'org.keycloak', name: 'keycloak-admin-client', version: '19.0.1') {
        because 'CVE. This constraint can be removed when this dependency is upgraded in com.ritense.valtimo:keycloak-iam ' +
                'See  https://nvd.nist.gov/vuln/detail/CVE-2022-1245'
    }
}
```

PR #282, opened 2022-08-18
feature 36775 - Added a new request param to /api/plugin/configuration

Rick-Ritense:
Using this parameter the configurations can be filtered to only include plugin configurations with actions

### DIFF
--- a/plugin/src/main/kotlin/com/ritense/plugin/domain/ActivityType.kt
+++ b/plugin/src/main/kotlin/com/ritense/plugin/domain/ActivityType.kt
@@ -16,22 +16,13 @@
 
 package com.ritense.plugin.domain
 
-import mu.KotlinLogging
-
-enum class ActivityType {
-    SERVICE_TASK;
+enum class ActivityType(
+    val bpmnModelValue: String
+) {
+    SERVICE_TASK("bpmn:ServiceTask");
 
     companion object {
-        fun from(bpmnTask: String?): ActivityType? {
-            return when(bpmnTask) {
-                "bpmn:ServiceTask" -> SERVICE_TASK
-                else -> {
-                    logger.error { "Unable to find an ActivityType for $bpmnTask" }
-                    null
-                }
-            }
-        }
-
-        val logger = KotlinLogging.logger {}
+        private val mapping = values().associateBy(ActivityType::bpmnModelValue)
+        fun fromValue(value:String) = mapping[value] ?: error("Can't find ActivityType with value $value")
     }
 }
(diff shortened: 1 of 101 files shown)


plugin/src/main/kotlin/com/ritense/plugin/web/rest/PluginConfigurationResource.kt
```diff
@@ -42,12 +41,26 @@ class PluginConfigurationResource(
 ) {
 
     @GetMapping(value = ["/configuration"], produces = [MediaType.APPLICATION_JSON_VALUE])
-    fun getPluginDefinitions(@RequestParam("category") category: String?): ResponseEntity<List<PluginConfigurationDto>> {
+    fun getPluginDefinitions(@RequestParam("category") category: String?,
+                             @RequestParam("includeActionless", defaultValue = "true") includeActionless: Boolean)
```
ivo-ritense, line 45: i think this should be a filter on the type of action instead

marijnritense: Agreed, when a service task is clicked, we should only show configurations with actions that work for service tasks.

plugin/src/main/kotlin/com/ritense/plugin/service/PluginConfigurationSearchService.kt
```diff
@@ -0,0 +1,66 @@
...
+import com.ritense.plugin.domain.PluginDefinition
+import javax.persistence.EntityManager
+import javax.persistence.TypedQuery
+import javax.persistence.criteria.CriteriaBuilder
+import javax.persistence.criteria.Join
+import javax.persistence.criteria.JoinType
+import javax.persistence.criteria.Predicate
+import javax.persistence.criteria.Root
+
+class PluginConfigurationSearchService(
```
marijnritense, line 15: I know I wrote this, but sanity check: Is this actually a service, or a repository?

Rick-Ritense: Good quesion. It creates a query so it should be a repository

plugin/src/main/kotlin/com/ritense/plugin/domain/ActivityType.kt
```diff
@@ -16,6 +16,22 @@
...
 package com.ritense.plugin.domain
 
+import mu.KotlinLogging
+
 enum class ActivityType {
-    SERVICE_TASK
-}
+    SERVICE_TASK;
+
+    companion object {
```
ivo-ritense, line 24: if we're doing this i'd rather have the corresponding key be part of the enum, instead of having another method having that knowledge. so something like this:

```
enum class ActivityType(
    val bpmnModelValue: String
) {

SERVICE_TASK("bpmn:ServiceTask")
```

Rick-Ritense: I'll have a look

ivo-ritense (approved):

PR #284, opened 2022-08-19
Fixed spelling error in variable name

ivo-ritense:


mritense (approved):

openzaak/src/main/kotlin/com/ritense/openzaak/service/impl/model/zaak/Zaak.kt
```diff
@@ -55,7 +55,7 @@ data class Zaak(
     val archiefstatus: String,
     val archiefactiedatum: String?,
     val resultaat: URI?,
-    var resulaatOmschrijving: String?
+    var resultaatOmschrijving: String?
```
marijnritense, line 58: This looks like it is a breaking change for impl teams that use this value.

PR #287, opened 2022-08-24
added isActive to CamundaProcessJsonSchemaDocumentInstance

sirolf2009:
Added a formula to check if a process connected to a document is in status ACTIVE.
To be used for sorting processes in the front end.

### DIFF
--- a/process-document/src/main/java/com/ritense/processdocument/domain/impl/CamundaProcessJsonSchemaDocumentInstance.java
+++ b/process-document/src/main/java/com/ritense/processdocument/domain/impl/CamundaProcessJsonSchemaDocumentInstance.java
@@ -17,7 +17,9 @@
 package com.ritense.processdocument.domain.impl;
 
 import com.fasterxml.jackson.annotation.JsonIgnore;
+import com.fasterxml.jackson.annotation.JsonProperty;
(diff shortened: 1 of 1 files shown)


process-document/src/main/java/com/ritense/processdocument/domain/impl/CamundaProcessJsonSchemaDocumentInstance.java
```diff
@@ -38,6 +40,14 @@
     @Column(name = "process_name", columnDefinition = "VARCHAR(255)")
     private String processName;
 
+    @Formula("( " +
```
Tom-Ritense, line 43: Iam missing the integration test that asserts this value is set please update existing or add new testcase.

Tom-Ritense (changes requested):

process-document/src/main/java/com/ritense/processdocument/domain/impl/CamundaProcessJsonSchemaDocumentInstance.java
sirolf2009: I have switched to a better method for querying statusses and expanded an integration test with some assertions

Tom-Ritense (approved):

PR #295, opened 2022-09-06
Load prefilled form for zaakobject

ivo-ritense:


app/gzac/src/main/resources/config/form/Boom.editform.json
```diff
@@ -0,0 +1,117 @@
+{
```
marijnritense, line 1: Maybe ignore case, so this can be lowercase?

Klaas-Ritense (approved):

marijnritense (approved):

PR #297, opened 2022-09-07
Delete process link -endpoint

Klaas-Ritense:


plugin/src/test/kotlin/com/ritense/plugin/web/rest/PluginProcessLinkResourceTest.kt
```diff
@@ -162,4 +163,15 @@ internal class PluginProcessLinkResourceTest {
 
         verify(pluginService).updateProcessLink(pluginProcessLinkDto)
     }
-}
\ No newline at end of file
+
+    @Test
+    fun `should delete plugin process link`() {
```
marijnritense, line 168: What happens when a process link that does not exist is deleted?

plugin/src/main/kotlin/com/ritense/plugin/web/rest/PluginProcessLinkResource.kt
```diff
@@ -64,4 +67,13 @@ class PluginProcessLinkResource(
...
-}
\ No newline at end of file
+
+    @DeleteMapping("/{processLinkId}")
+    fun deleteProcessLink(
+        @PathVariable(name = "processLinkId") processLinkId: UUID
+    ): ResponseEntity<Unit> {
+        pluginService.deleteProcessLink(processLinkId)
+
+        return ResponseEntity.status(HttpStatus.NO_CONTENT).build()
```
ivo-ritense, line 77: What happens when a process link that does not exist is deleted?

ivo-ritense: i guess we want to have a delete endpoint be idempotent?

marijnritense (approved):

PR #302, opened 2022-09-13
#37108 system processes

Klaas-Ritense:


core/src/main/java/com/ritense/valtimo/service/ProcessPropertyService.java
```diff
@@ -0,0 +1,69 @@
...
+            throw new RuntimeException("Failed to find properties for process with key: " + processDefinitionKey);
+        }
+        return processProperties.isSystemProcess();
+    }
+
+    public boolean isReadOnly(String processDefinitionKey) {
+        return !valtimoProperties.getProcess().isSystemProcessUpdatable() && isSystemProcess(processDefinitionKey);
+    }
+
+    public boolean isSystemProcessById(String processDefinitionId) {
```
theo-ritense, line 51: This should be moved above isSystemProcess. Method order should follow the code as much as possible for readability.

core/src/main/java/com/ritense/valtimo/service/ProcessPropertyService.java
```diff
@@ -0,0 +1,69 @@
...
+        return isSystemProcess(getProcessDefinitionKeyById(processDefinitionId));
+    }
+
+    public boolean isReadOnlyById(String processDefinitionId) {
```
theo-ritense, line 55: This should be moved above isReadOnly

core/src/test/java/com/ritense/valtimo/web/rest/ProcessResourceIntTest.java
```diff
@@ -0,0 +1,78 @@
...
+
+        mockMvc.perform(get("/api/process/definition/{processDefinitionId}/xml", processDefinition.getId())
+                .accept(APPLICATION_JSON_VALUE))
+            .andDo(print())
+            .andExpect(status().isOk())
+            .andExpect(jsonPath("$.id").value(processDefinition.getId()))
+            .andExpect(jsonPath("$.bpmn20Xml").isNotEmpty())
+            .andExpect(jsonPath("$.readOnly").value(true))
+            .andExpect(jsonPath("$.systemProcess").value(true));
+    }
```
theo-ritense, line 76: I would expect a test to see if an update to a read-only processdefinition throws an error.

core/src/main/java/com/ritense/valtimo/service/ProcessPropertyService.java
Klaas-Ritense: changed

core/src/main/java/com/ritense/valtimo/service/ProcessPropertyService.java
Klaas-Ritense: changed

core/src/test/java/com/ritense/valtimo/web/rest/ProcessResourceIntTest.java
Klaas-Ritense: I added a test. But not for your use-case because I found an issue: 
The FE calls this endpoint for deploying a process:  `/camunda-rest/engine/default/deployment/create`. This is a camunda endpoint and overriding it is a huge pain. Another solution is to make a new endpoint in Valtimo, which would be a breaking change. 

I didn't do either. So for now its only FE that makes sure that people dont deploy a readOnly process. and leave this problem for later :P

marijnritense: I'll create a story to cover this from a BE perspective as well in major 10.

Rick-Ritense (approved):

PR #303, opened 2022-09-13
feature 37202 - Upload document process

Rick-Ritense:
This new process is marked as a system process


Klaas-Ritense (approved):

Klaas-Ritense (changes requested):
This location isn't going to work for

app/gzac/src/main/resources/bpmn/upload-document.bpmn
```diff
@@ -0,0 +1,55 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
Klaas-Ritense, line 1: The BPMN location is now in: `app/gzac/src/main/resources/bpmn/upload-document.bpmn`. This wont work for implementation projects since they don't use `app/gzac/...`

A better location would be the `core` module.

Also maybe a problem for later: This process should run out-of-the-box. I'm not sure how we're going to add the process-links out-of-the-box. Maybe the `expression="${true}"` should change in the future

Rick-Ritense: you are right about the location. I'll look into that

About the configuration. The main story says this:

```
This BPMN process handles the transfer of the file to the Document API, as well as the linking of the Document to the Zaak
These two actions are process-links and can be configured by the GZAC administrator
```

So it should not be working out of the box ... there is always some configuration needed by an administrator

ThomasRitense: Since this process has ZGW specific tasks, may I suggest putting it in the same module as the Documenten API plugin?

marijnritense: Wouldn't the zaken API make more sense then? That one has a dependency on the documenten API module as well, and both have to be configured here.

If we want to add a way to configure process links out of the box, then we have to allow them to not be connected to a plugin configuration yet (but do allow them to be associated with a plugin definition? That way the process link can mostly be configured). @ThomasRitense Does this need to work fully out of the box?

PR #305, opened 2022-09-14
Include formDefinition in FormFieldDataResolver.get method

maarten-ritense:
Added 'get' to FormFieldDataResolver which includes the formDefinition JsonNode parameter, marked the other 'get' a Deprecated.

### DIFF
--- a/contract/src/main/java/com/ritense/valtimo/contract/form/FormFieldDataResolver.java
+++ b/contract/src/main/java/com/ritense/valtimo/contract/form/FormFieldDataResolver.java
@@ -36,11 +36,9 @@ public interface FormFieldDataResolver {
     }
 
     default Map<String, Object> get(
-        String documentDefinitionName,
-        UUID documentId,
-        FormFieldDataResolverProperties additionalProperties,
+        DataResolvingContext dataResolvingContext,
         String... varNames
     ) {
-        return get(documentDefinitionName, documentId, varNames);
+        return get(dataResolvingContext.getDocumentDefinitionName(), dataResolvingContext.getDocumentId(), varNames);
     }
 }
--- a/contract/src/main/java/com/ritense/valtimo/contract/form/FormFieldDataResolver.java
+++ b/contract/src/main/java/com/ritense/valtimo/contract/form/FormFieldDataResolver.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2015-2020 Ritense BV, the Netherlands.
+ * Copyright 2015-2022 Ritense BV, the Netherlands.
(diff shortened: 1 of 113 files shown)


Tom-Ritense (approved):

contract/src/main/java/com/ritense/valtimo/contract/form/FormFieldDataResolver.java
```diff
@@ -30,6 +32,8 @@ default boolean supports(String externalFormFieldType) {
         return supports(ExternalFormFieldType.fromKey(externalFormFieldType));
     }
 
+    @Deprecated(forRemoval = true, since = "9.21")
     Map<String, Object> get(String documentDefinitionName, UUID documentId, String... varNames);
 
+    Map<String, Object> get(String documentDefinitionName, UUID documentId, JsonNode formDefinition, String... varNames);
```
marijnritense, line 38: If you give this method a default implementation (e.g. just call the old get) it won't be a breaking change.

Tom-Ritense (approved):

contract/src/main/java/com/ritense/valtimo/contract/form/FormFieldDataResolver.java
```diff
@@ -30,6 +30,15 @@ default boolean supports(String externalFormFieldType) {
         return supports(ExternalFormFieldType.fromKey(externalFormFieldType));
     }
 
-    Map<String, Object> get(String documentDefinitionName, UUID documentId, String... varNames);
+    @Deprecated(forRemoval = true, since = "9.21")
+    default Map<String, Object> get(String documentDefinitionName, UUID documentId, String... varNames) {
```
marijnritense, line 34: What I meant was for the new method to be default, not this one. So:

```
    default Map<String, Object> get(
        String documentDefinitionName,
        UUID documentId,
        FormFieldDataResolverProperties additionalProperties,
        String... varNames
    ) {
        return get(documentDefinitionName, documentId, varNames);
    }
```

Rick-Ritense: doesn't that make the code breaking when you change the default

I think you should change the default when the old method is removed in the future

ivo-ritense: The default would be to ensure backwards compatibility, the intent is still to override it in your implementation.

marijnritense (approved):

ivo-ritense (approved):

PR #308, opened 2022-09-16
#37201 Start BPMN upload process after upload

Klaas-Ritense:


resource/temporary-resource-storage/src/main/kotlin/com/ritense/resource/domain/MetadataType.kt
```diff
@@ -1,7 +1,9 @@
 package com.ritense.resource.domain
 
-enum class MetadataType {
-    FILE_NAME,
-    FILE_PATH,
-    CONTENT_TYPE,
+enum class MetadataType(val value: String) {
```
theo-ritense, line 3: It's not a value though. It's a fieldName, key or propertyName.

Klaas-Ritense: renamed

PR #310, opened 2022-09-16
#37205 Endpoint to check whether an upload process has been linked to…

Klaas-Ritense:
… a case definition

### DIFF
--- a/zgw/zaken-api/src/main/kotlin/com/ritense/zakenapi/uploadprocess/UploadProcessAutoConfiguration.kt
+++ b/zgw/zaken-api/src/main/kotlin/com/ritense/zakenapi/uploadprocess/UploadProcessAutoConfiguration.kt
@@ -1,0 +1,56 @@
+/*
+ * Copyright 2015-2022 Ritense BV, the Netherlands.
+ *
+ * Licensed under EUPL, Version 1.2 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * https://joinup.ec.europa.eu/collection/eupl/eupl-text-eupl-12
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" basis,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.ritense.zakenapi.uploadprocess
+
+import com.ritense.processdocument.service.ProcessDocumentAssociationService
+import com.ritense.processdocument.service.ProcessDocumentService
+import com.ritense.resource.service.TemporaryResourceStorageService
+import org.springframework.boot.autoconfigure.condition.ConditionalOnMissingBean
+import org.springframework.context.annotation.Bean
+import org.springframework.context.annotation.Configuration
+import org.springframework.core.annotation.Order
+
+@Configuration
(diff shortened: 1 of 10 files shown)


zgw/zaken-api/src/main/kotlin/com/ritense/zakenapi/uploadprocess/UploadProcessResource.kt
```diff
@@ -0,0 +1,49 @@
...
+        val processDocumentDefinitions = processDocumentAssociationService.findAllProcessDocumentDefinitions(
+            CamundaProcessDefinitionKey(
+                UPLOAD_DOCUMENT_PROCESS_DEFINITION_KEY
+            )
+        )
+
+        val processCaseLinkExists = processDocumentDefinitions.any {
+            caseDefinitionKey == it.processDocumentDefinitionId().documentDefinitionId().name()
+        }
+        return ResponseEntity.ok(CheckLinkResponse(processCaseLinkExists))
```
theo-ritense, line 47: What if we rename the mapping to `/case/{caseDefinitionKey}/link` and respond with either a 200 or 404?

zgw/zaken-api/src/main/kotlin/com/ritense/zakenapi/uploadprocess/UploadProcessAutoConfiguration.kt
```diff
@@ -0,0 +1,56 @@
...
+
+    @Bean
+    @ConditionalOnMissingBean(UploadProcessResource::class)
+    fun uploadProcessResource(
+        processDocumentAssociationService: ProcessDocumentAssociationService
+    ): UploadProcessResource {
+        return UploadProcessResource(processDocumentAssociationService)
+    }
+
+    @Order(360)
```
theo-ritense, line 50: Why is the order needed?

zgw/zaken-api/src/main/kotlin/com/ritense/zakenapi/uploadprocess/UploadProcessResource.kt
Klaas-Ritense: @mritense Would a 404 response work for the FE?

zgw/zaken-api/src/main/kotlin/com/ritense/zakenapi/uploadprocess/UploadProcessAutoConfiguration.kt
Klaas-Ritense: Im not sure. We have the @Order above every HttpSecurityConfigurer and the order does matter. See:
https://github.com/valtimo-platform/valtimo-backend-libraries/blob/development/core/src/main/java/com/ritense/valtimo/autoconfigure/HttpSecurityAutoConfiguration.java#L109

But I dont know how to decide which order is the correct one

zgw/zaken-api/src/main/kotlin/com/ritense/zakenapi/uploadprocess/UploadProcessResource.kt
Klaas-Ritense: I just asked Maxim and the FE recognizes the 404 response as an error and will show toastr error popup. But, the toastr error can be intercepted. 

Still it more easy for the FE to respond with a boolean

theo-ritense: ok

theo-ritense (approved):

PR #318, opened 2022-09-26
#37224 endpoint for uploading form files to local file storage

Klaas-Ritense:


process-document/src/main/java/com/ritense/processdocument/service/DocumentDefinitionProcessLinkService.java
```diff
@@ -21,13 +21,14 @@
 import com.ritense.processdocument.domain.impl.request.DocumentDefinitionProcessLinkResponse;
 import com.ritense.processdocument.domain.impl.request.DocumentDefinitionProcessRequest;
 
+import java.util.List;
 import java.util.Optional;
 
 public interface DocumentDefinitionProcessLinkService {
 
-    DocumentDefinitionProcess getDocumentDefinitionProcess(String documentDefinitionName);
+    List<DocumentDefinitionProcess> getDocumentDefinitionProcess(String documentDefinitionName);
```
ivo-ritense, line 29: This is a breaking change. Is it possible to create this method next to the existing one?

process-document/src/main/java/com/ritense/processdocument/service/DocumentDefinitionProcessLinkService.java
```diff
@@ -21,13 +21,14 @@
...
 
-    Optional<DocumentDefinitionProcessLink> getDocumentDefinitionProcessLink(String documentDefinitionName);
+    Optional<DocumentDefinitionProcessLink> getDocumentDefinitionProcessLink(String documentDefinitionName, String type);
```
marijnritense, line 31: shouldn't this be covered by tests?

process-document/src/main/java/com/ritense/processdocument/service/impl/DocumentDefinitionProcessLinkServiceImpl.java
```diff
@@ -41,24 +43,22 @@ public DocumentDefinitionProcessLinkServiceImpl(DocumentDefinitionProcessLinkRep
...
-    public DocumentDefinitionProcess getDocumentDefinitionProcess(String documentDefinitionName) {
-        var link= documentDefinitionProcessLinkRepository.findByIdDocumentDefinitionName(documentDefinitionName);
+    public List<DocumentDefinitionProcess> getDocumentDefinitionProcess(String documentDefinitionName) {
+        var links = documentDefinitionProcessLinkRepository.findAllByIdDocumentDefinitionName(documentDefinitionName);
 
-        if (link.isPresent()) {
+        return links.stream().map(link -> {
             var processDefinition = repositoryService.createProcessDefinitionQuery()
-                .processDefinitionKey(link.get().getId().getProcessDefinitionKey())
+                .processDefinitionKey(links.get(0).getId().getProcessDefinitionKey())
```
ivo-ritense, line 51: Aren't you always still getting the first entry in the list this way? Then what's the point of getting a list?

resource/temporary-resource-storage/src/main/kotlin/com/ritense/resource/service/TemporaryResourceStorageService.kt
```diff
@@ -62,14 +65,31 @@ class TemporaryResourceStorageService {
...
+        }
+        val fileNumber = resourceUuid.leastSignificantBits.toULong()
+        return Path.of(TEMP_DIR.pathString, "temporaryResourceMetadata$fileNumber.json")
+    }
+
+    private fun getResourceIdFromMetaDataFile(tempFile: Path): String {
+        if (!tempFile.nameWithoutExtension.startsWith("temporaryResourceMetadata")) {
+            throw IllegalArgumentException("File is not a temporaryResourceMetadata file: $tempFile")
+        }
+        val fileNumber = tempFile.nameWithoutExtension.substring("temporaryResourceMetadata".length).toULong().toLong()
```
ivo-ritense, line 89: I'd prefer to have this not be a UUID based on the randomly generated name. The Java documentation says
```
The details as to how the name of the file is constructed is implementation dependent and therefore not specified. Where possible the prefix and suffix are used to construct candidate names in the same manner as the File.createTempFile(String, String, File) method.
```
So i think we can't 100% expect this to be a long. I think we should just return the file name (not the path).

process-document/src/main/java/com/ritense/processdocument/service/DocumentDefinitionProcessLinkService.java
Klaas-Ritense: Changed

process-document/src/main/java/com/ritense/processdocument/service/DocumentDefinitionProcessLinkService.java
Klaas-Ritense: Added

process-document/src/main/java/com/ritense/processdocument/service/impl/DocumentDefinitionProcessLinkServiceImpl.java
Klaas-Ritense: Changed

resource/temporary-resource-storage/src/main/kotlin/com/ritense/resource/service/TemporaryResourceStorageService.kt
Klaas-Ritense: Changed

ivo-ritense (approved):

PR #327, opened 2022-10-03
bug: FormIoFormDefinition::extractNodeValue incorrectly parsing types

anna-ritense:
* modified type checks to be more specific and added handling for number and boolean nodes
* updated tests and test form definition
* made private constructor protected as per a jpa requirement

### DIFF
--- a/form/src/main/java/com/ritense/form/domain/FormIoFormDefinition.java
+++ b/form/src/main/java/com/ritense/form/domain/FormIoFormDefinition.java
@@ -21,9 +21,13 @@ import com.fasterxml.jackson.core.JsonPointer;
 import com.fasterxml.jackson.databind.JsonNode;
 import com.fasterxml.jackson.databind.node.ArrayNode;
 import com.fasterxml.jackson.databind.node.JsonNodeFactory;
+import com.fasterxml.jackson.databind.node.JsonNodeType;
 import com.fasterxml.jackson.databind.node.ObjectNode;
 import com.fasterxml.jackson.databind.node.TextNode;
 import com.ritense.form.domain.event.FormRegisteredEvent;
+import static com.ritense.valtimo.contract.utils.AssertionConcern.assertArgumentLength;
+import static com.ritense.valtimo.contract.utils.AssertionConcern.assertArgumentNotNull;
+import static com.ritense.valtimo.contract.utils.AssertionConcern.assertStateTrue;
 import org.hibernate.annotations.Type;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
@@ -45,10 +49,6 @@ import java.util.List;
 import java.util.Map;
 import java.util.Optional;
 import java.util.UUID;
-
-import static com.ritense.valtimo.contract.utils.AssertionConcern.assertArgumentLength;
-import static com.ritense.valtimo.contract.utils.AssertionConcern.assertArgumentNotNull;
-import static com.ritense.valtimo.contract.utils.AssertionConcern.assertStateTrue;
 
 @Entity
 @Table(name = "form_io_form_definition")
@@ -104,7 +104,7 @@ public class FormIoFormDefinition extends AbstractAggregateRoot<FormIoFormDefini
         registerEvent(new FormRegisteredEvent(id, name));
     }
 
-    private FormIoFormDefinition() {
+    protected FormIoFormDefinition() {
     }
 
     public void setReadOnly(Boolean value) {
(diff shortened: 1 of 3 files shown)


form/src/main/java/com/ritense/form/domain/FormIoFormDefinition.java
```diff
@@ -408,9 +408,14 @@ private static List<ArrayNode> getComponents(JsonNode formDefinition) {
...
+        if (nodeType == JsonNodeType.STRING) {
             return node.textValue();
-        } else if (node.isArray()) {
+        } else if (nodeType == JsonNodeType.NUMBER) {
+            return node.numberValue();
+        } else if (nodeType == JsonNodeType.BOOLEAN) {
+            return node.booleanValue();
+        } else if (nodeType == JsonNodeType.ARRAY) {
             List<String> values = new ArrayList<>();
             node.forEach(childNode -> values.add(childNode.textValue()));
```
theo-ritense, line 420: shouldn't this be some recursive method call, like
`node.forEach(childNode -> values.add(extractNodeValue(childNode)));`

?

anna-ritense: > shouldn't this be some recursive method call, like
This is mainly a bugfix and not a functional improvement. If you think this should be made recursive instead of letting us just update the whole array at once, then that can be discussed separately.

theo-ritense: Just looking at the code, the bug still exists when trying to access an array node with non-text values. So the bug would not be fully fixed as it is.

anna-ritense: Then this is a bigger issue in general. What if the array contains non-value nodes (objects)?
Recursion will add a lot of complication here and we only really should need to overwrite the whole array in the process variable.

I'm welcome to suggestions here, because we can not assume the type of every element in the array.

anna-ritense: >  It aims to resolve the issues the tribe is currently experiencing with one of their implementations 

It aims to solve an issue with pv. that is used in two implementations. This PR aims to solve valueNode parsing issues in this method, not containerNodes. containerNodes should've never been parsed in here to begin with since they need to be approached differently.

> I think we either need to decline the PR and figure out what exactly we want to do here, or look at the PR for what it is 

Both of these are fine with me as long as this issue gets resources assigned to it and not just put on the backlog.

form/src/main/java/com/ritense/form/domain/FormIoFormDefinition.java
```diff
@@ -408,9 +408,14 @@ private static List<ArrayNode> getComponents(JsonNode formDefinition) {
     }
 
     private Object extractNodeValue(JsonNode node) {
-        if (node.isValueNode()) {
+        var nodeType = node.getNodeType();
+        if (nodeType == JsonNodeType.STRING) {
             return node.textValue();
-        } else if (node.isArray()) {
+        } else if (nodeType == JsonNodeType.NUMBER) {
+            return node.numberValue();
```
Klaas-Ritense, line 415: This is a bit of a breaking change. Processes that used to have a process variable with type 'string' can now get the same process variable with type 'boolean' or 'number'.

So this PR should go to major 10. We have a branch for that, can you change the destination to the major 10 branch?

form/src/main/java/com/ritense/form/domain/FormIoFormDefinition.java
Klaas-Ritense: This PR does solve an issue, so thats good. Lets figure out what to do with the array some other time.

form/src/main/java/com/ritense/form/domain/FormIoFormDefinition.java
anna-ritense: > This is a bit of a breaking change. Processes that used to have a process variable with type 'string' can now get the same process variable with type 'boolean' or 'number'.

Before this change this method would make every valueNode other than TextNode a null value. We always had the ability to submit what ever we wanted in form-io - that means that even before this fix we could've made a previously string value into a integer, thus the end result becoming null in the processvariable in camunda.
The only thing I now did is that the returned value aligns with it's nodeType - how is this breaking?

> So this PR should go to major 10. We have a branch for that, can you change the destination to the major 10 branch?

I need the valueNode types fixed asap, for containerNodes you can take this into major 10 if you like. If you don't want to accept this PR on the premise that it only fixes valueNodes in the extractNodeValue method, then close the PR and I'll make a story for team Valtimo instead.

Klaas-Ritense: A right, values of other types become null

Klaas-Ritense (approved):

PR #333, opened 2022-10-10
#37531 API for adding notes

Klaas-Ritense:


notes/src/main/kotlin/com/ritense/note/domain/Note.kt
```diff
@@ -0,0 +1,55 @@
...
+
+    @Column(name = "content", columnDefinition = "CLOB", nullable = false)
+    val content: String,
+
+    @Column(name = "document_id", nullable = false)
+    val documentId: UUID,
+) {
+    constructor(documentId: JsonSchemaDocumentId, content: String) : this(
+        UUID.randomUUID(),
+        SecurityUtils.getCurrentUserLogin(),
```
ivo-ritense, line 50: I think we might want the same stored id + users name here that we use for case assignee.

notes/src/main/kotlin/com/ritense/note/web/rest/NoteResource.kt
```diff
@@ -0,0 +1,51 @@
...
+) {
+
+    @PostMapping
+    fun createNote(
+        @PathVariable(name = "documentId") documentId: UUID,
+        @RequestBody noteDto: NoteCreateRequestDto
+    ): ResponseEntity<NoteResponseDto> {
+
+        val note = noteService.createNote(
+            JsonSchemaDocumentId.existingId(documentId),
```
ivo-ritense, line 45: Do we need to check access of the user to this type of case?

notes/src/main/kotlin/com/ritense/note/domain/Note.kt
Klaas-Ritense: I'm going to change the separate first name and last name to one field: fullname. I suspect that thats better for FE

PR #340, opened 2022-10-14
Added max gradle workers property.

ThomasRitense:


gradle.properties
```diff
@@ -24,6 +24,7 @@ namedVolumeNameTest=valtimo-test-fixtures
 org.gradle.daemon=true
 org.gradle.jvmargs=-Xmx6g -XX:MaxPermSize=2048m -XX:+HeapDumpOnOutOfMemoryError
 org.gradle.parallel=true
+org.gradle.workers.max=3
```
theo-ritense, line 27: Please note that the default is 'the number of CPU processors/cores'. This setting could become a problem on smaller machines. Doubt that any developer machine has less than 4 cores, but an EC2 machine might.

theo-ritense (approved):

gradle.properties
ThomasRitense: Thanks for the note, indeed not fixing as we configure the CI/CD instances ourselves.

PR #343, opened 2022-10-17
Make form flow available for postgres

ivo-ritense:
Introduces a new setup used for the ObjectMapper used by the hibernate json types in order to prevent the incompatible converter to break when used with PostgreSQL.


form-flow/src/main/kotlin/com/ritense/formflow/service/ObjectMapperConfigurer.kt
```diff
@@ -18,26 +18,25 @@ package com.ritense.formflow.service
...
-                ?.forEach {
-                    registerSubtypes(it)
-                }
-        }
-
-    fun get(): ObjectMapper {
-        return mapper
+    @EventListener(ApplicationStartedEvent::class)
+    fun configure() {
+        objectMapper.registerKotlinModule()
```
theo-ritense, line 33: You're actually changing the application-wide objectmapper like this, as the `registerKotlinModule` is mutating `this`. Not sure if that's the desired effect.

ivo-ritense: That is what i intended to do. Maybe the kotlin module itself doesn't have to be in this configuration.

marijnritense (approved):

PR #345, opened 2022-10-17
Feature/exact plugin

nielserkamp-ritense:


nielserkamp-ritense:
@marijnritense @ivo-ritense 

Currently the plugin uses the client ID and secret as filled in the plugin, however I am wondering if it would not be better to make this 'part' of the application properties. As you would still need to authorize with a user for an access token and refresh token. 

This also removes the need to fill in a client id and secret when setting up the configuration for the plugin, which makes it more user friendly.

Let me know what you think.

ivo-ritense (approved):

PR #349, opened 2022-10-18
#37531 get current user that has a full name

Klaas-Ritense:
The existing `CurrentUserRepository.getCurrentUser()` doesn't return the name of the user. Only the email.

The new `UserManagementService.getCurrentUser()` does return the name of the user including many other properties

### DIFF
--- a/keycloak-iam/src/main/java/com/valtimo/keycloak/service/KeycloakUserManagementService.java
+++ b/keycloak-iam/src/main/java/com/valtimo/keycloak/service/KeycloakUserManagementService.java
@@ -22,6 +22,7 @@ import com.ritense.valtimo.contract.authentication.UserNotFoundException;
 import com.ritense.valtimo.contract.authentication.model.SearchByUserGroupsCriteria;
 import com.ritense.valtimo.contract.authentication.model.ValtimoUser;
 import com.ritense.valtimo.contract.authentication.model.ValtimoUserBuilder;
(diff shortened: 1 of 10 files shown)


keycloak-iam/src/main/java/com/valtimo/keycloak/service/KeycloakUserManagementService.java
```diff
@@ -163,6 +166,17 @@ public List<ManageableUser> findByRoles(SearchByUserGroupsCriteria groupsCriteri
             .collect(Collectors.toList());
     }
 
+    @Override
+    public ManageableUser getCurrentUser() {
```
ivo-ritense, line 170: This probably works, but why don't we find the user by id? I'm not sure if a user always has an email, but they should always have an id.

Klaas-Ritense: Because this is for keycloak. Keycloak doesnt have a user ID. Keycloak puts only the email into the Authentication object. See:
https://github.com/valtimo-platform/valtimo-backend-libraries/blob/development/keycloak-iam/src/main/java/com/valtimo/keycloak/security/jwt/authentication/KeycloakTokenAuthenticator.java#L83

ivo-ritense: Not sure that's the best idea, but for now it's fine then.

PR #357, opened 2022-10-24
feature 37577 - Added SearchField entity and ability to store an entity

Rick-Ritense:


document/src/main/resources/config/liquibase/changelog/202221021-add-search-field-table.xml
```diff
@@ -0,0 +1,16 @@
+<databaseChangeLog xmlns="http://www.liquibase.org/xml/ns/dbchangelog"
```
Klaas-Ritense, line 1: Seems like we have some duplicate work:
https://github.com/valtimo-platform/valtimo-backend-libraries/pull/355/files#diff-4f74a6f262610c5200d54201b733ae0e34f02717756242a46f506ff607adc8a3

Klaas-Ritense (approved):

PR #359, opened 2022-10-26
#37575 Autodeployment of case search fields

Klaas-Ritense:


document/src/main/java/com/ritense/document/domain/impl/searchfield/SearchField.java
```diff
@@ -16,8 +16,8 @@
 
 package com.ritense.document.domain.impl.searchfield;
 
-import com.fasterxml.jackson.annotation.JsonValue;
 import org.springframework.data.domain.Persistable;
+
```
luis-soares-ritense, line 20: entity properties should have the same constraints according to the constraints added to the properties made in liquibase migration(document/src/main/resources/config/liquibase/changelog/20221021-add-search-field-table.xml)

Klaas-Ritense: I've added them

luis-soares-ritense (approved):

PR #360, opened 2022-10-26
Feature/37580 api endpoint to edit case search field

luis-soares-ritense:


document/src/main/java/com/ritense/document/web/rest/impl/SearchFieldResource.java
```diff
@@ -53,4 +54,15 @@ public ResponseEntity<List<SearchField>> getSearchField(
         @PathVariable String documentDefinitionName) {
         return ResponseEntity.ok(searchFieldService.getSearchFields(documentDefinitionName));
     }
+
+    @Override
+    @PutMapping("/v1/document-search/{documentDefinitionName}/fields")
```
Klaas-Ritense, line 59: This endpoint should also be added to the SecurityConfiguration. See:
https://github.com/valtimo-platform/valtimo-backend-libraries/blob/development/document/src/main/java/com/ritense/document/security/config/DocumentSearchHttpSecurityConfigurer.java#L33

luis-soares-ritense: added

document/src/main/java/com/ritense/document/web/rest/DocumentSearchFields.java
```diff
@@ -26,4 +28,8 @@ public interface DocumentSearchFields {
     ResponseEntity<Void> addSearchField(String documentDefinitionName, SearchField searchField);
 
     ResponseEntity<List<SearchField>> getSearchField(String documentDefinitionName);
+
+    @PutMapping("/v1/document-search/{documentDefinitionName}/fields")
```
Klaas-Ritense, line 32: Having the annotation twice is a bit much. This one can be removed

document/src/main/java/com/ritense/document/web/rest/impl/SearchFieldResource.java
```diff
@@ -53,4 +54,15 @@ public ResponseEntity<List<SearchField>> getSearchField(
...
+    public ResponseEntity<List<SearchField>> updateSearchField(
+            @RequestBody SearchField searchField) {
+        if(searchField.getId() == null){
```
Klaas-Ritense, line 62: Thet frontend might not send back the 'id' property. Because the frontend isn't even supposed to rget the 'id'. They work with the 'key' property

document/src/main/java/com/ritense/document/service/SearchFieldService.java
```diff
@@ -39,4 +39,8 @@ public void addSearchField(String documentDefinitionName, SearchField searchFiel
     public List<SearchField> getSearchFields(String documentDefinitionName) {
         return searchFieldRepository.findAllByIdDocumentDefinitionName(documentDefinitionName);
     }
+
+    public void updateSearchFields(SearchField searchField) {
+        searchFieldRepository.save(searchField);
```
Klaas-Ritense, line 44: This is one of the acceptance criteria:
```
When a field exists for the document definition with the key, the field is replaced.
```

document/src/main/java/com/ritense/document/web/rest/impl/SearchFieldResource.java
luis-soares-ritense: yeah i thought of that but i wasn't sure. will fix

document/src/main/java/com/ritense/document/service/SearchFieldService.java
luis-soares-ritense: was assuming the id.

document/src/main/java/com/ritense/document/web/rest/impl/SearchFieldResource.java
luis-soares-ritense: fixed

document/src/main/java/com/ritense/document/service/SearchFieldService.java
luis-soares-ritense: fixed i believe

Klaas-Ritense (approved):

PR #362, opened 2022-10-28
Feature/37581 api endpoint to delete case search field

luis-soares-ritense:


document/src/main/java/com/ritense/document/service/SearchFieldService.java
```diff
@@ -52,4 +52,9 @@ public void createSearchConfiguration(String documentDefinitionName, List<Search
             searchFieldRepository.saveAll(searchFields);
         }
     }
+
+    public void deleteSearchFields(String documentDefinitionName, String key) {
```
marijnritense, line 56: This says field*s* (plural), but shouldn't it be singular?

luis-soares-ritense: you're right. will fix

ivo-ritense (approved):

PR #363, opened 2022-10-31
Feature/disabled input forms filter

YouriDudock:


ThomasRitense:
@YouriDudock Should a disabled field be pre-filled?

anna-ritense:
@YouriDudock Does this take into account that "disabled" is a native formio component property with it's own functionality. In some forms we disable and enable components based on other form components, thus creating a situation where we can no longer submit those changes.
Example:
Form initialises a disabled textfield component and a radio button. Based on radio input we either want to let the user modify the data in the textfield or not. But now because the form definition says this textfield was disabled on form load, the user's changes will be discarded.

YouriDudock:
> @YouriDudock Does this take into account that "disabled" is a native formio component property with it's own functionality. In some forms we disable and enable components based on other form components, thus creating a situation where we can no longer submit those changes. Example: Form initialises a disabled textfield component and a radio button. Based on radio input we either want to let the user modify the data in the textfield or not. But now because the form definition says this textfield was disabled on form load, the user's changes will be discarded.

I would assume that the form definition always represents the newest form definition during submit, so any switches from disabled to enabled should be captured. In your example, the data should only be send if the radiobutton enables the textfield. 

This is just my opinion. Team @valtimo-platform/valtimo-product-team @valtimo-platform/valtimo should further specify requirements / constraints

YouriDudock:
> @YouriDudock Should a disabled field be pre-filled?

Good question. I think team @valtimo-platform/valtimo @valtimo-platform/valtimo-product-team should further specify requirement/constraints

anna-ritense:
> In your example, the data should only be send if the radiobutton enables the textfield.

This is not how the `disabled` property works in formio. If we want data to not be accepted, we set `"input": false`. The functionality of `"disabled": true` in formio is cosmetic/interactive. We still want the data of a disabled component to be accepted.

Another example:
We have a list of incomes and at the bottom of the list a disabled textfield component that the user can not modify, but can see the total being summed up. We still want to submit that total and it be accepted.

YouriDudock:
> > In your example, the data should only be send if the radiobutton enables the textfield.
> 
> This is not how the `disabled` property works in formio. If we want data to not be accepted, we set `"input": false`. The functionality of `"disabled": true` in formio is cosmetic/interactive. We still want the data of a disabled component to be accepted.
> 
> Another example: We have a list of incomes and at the bottom of the list a disabled textfield component that the user can not modify, but can see the total being summed up. We still want to submit that total and it be accepted.

This is an incorrect assumption on how Valtimo works. The function that looks for the 'input=false' is meant to detect input components. It was not meant for setting actual input components to false, so that the data won't be saved. This can be confirmed by looking at the acceptance criteria of what an input component is according to Valtimo:

`(isInputComponent(fieldNode) || isTextFieldComponent(fieldNode)) && !isButtonTypeComponent(fieldNode)`

Here we can see that for a text field that it is always an input component, thus it will be saved. It does not matter if the input was set to false.

Using disabled to filter a field during a submit is common and accepted practise for web. It is the default behavior for any HTML form submit. We should follow this standard and not make up our own versions.

However your example might be a real problem, we should definitely look into a proper solution for this. Which might already exists if we look into the standard.

Rick-Ritense (approved):

PR #365, opened 2022-10-31
Updated dependencies

ivo-ritense:


openzaak/src/main/kotlin/com/ritense/openzaak/form/OpenZaakFormFieldDataResolver.kt
```diff
@@ -23,13 +23,13 @@ import com.ritense.openzaak.service.impl.ZaakService
...
 import com.ritense.valtimo.contract.form.ExternalFormFieldType
 import com.ritense.valtimo.contract.form.FormFieldDataResolver
-import java.util.UUID
 
 class OpenZaakFormFieldDataResolver(
     private val zaakService: ZaakService,
     private val zaakInstanceLinkService: ZaakInstanceLinkService
 ) : FormFieldDataResolver {
 
+    @Deprecated("Deprecated in Java")
```
Rick-Ritense, line 32: I don't fully get what the message is supposed to say. It will be removed for java but will still be available for Kotlin?

The @Deprecated should also include the "since" attribute telling in what version the method became deprecated so we know when we can remove it

milestones/build.gradle
```diff
@@ -59,15 +59,15 @@ dependencies {
     implementation "org.springframework:spring-jdbc"
 
     // Mapstruct
-    implementation "org.mapstruct:mapstruct-jdk8:${mapstruct_version}"
-    annotationProcessor "org.mapstruct:mapstruct-processor:${mapstruct_version}"
+    implementation "org.mapstruct:mapstruct-jdk8:${mapstructVersion}"
+    annotationProcessor "org.mapstruct:mapstruct-processor:${mapstructVersion}"
```
Rick-Ritense, line 63: I get a compilation error while running all the tests on this part

```
D:\Workspace\valtimo\valtimo-backend-libraries\milestones\src\main\java\com\ritense\valtimo\milestones\autoconfigure\MilestonesAutoConfiguration.java:33: error: package org.mapstruct.factory does not exist
import org.mapstruct.factory.Mappers;
```

Rick-Ritense (approved):

openzaak/src/main/kotlin/com/ritense/openzaak/form/OpenZaakFormFieldDataResolver.kt
ivo-ritense: It's a new kotlin thing where you are required to mark methods as deprecated when the method is deprecated in the class or interface that defines it. I could change this, but it should be removed in the next story anyway.

Rick-Ritense: ok didn't know that. Thanks for the clarification

PR #372, opened 2022-11-02
Introduced springdoc for OpenAPI support.

ivo-ritense:


web/src/main/java/com/ritense/valtimo/web/autoconfigure/SwaggerAutoConfiguration.java
```diff
@@ -58,43 +50,35 @@ public class SwaggerAutoConfiguration {
      * @return the Swagger Springfox configuration
      */
     @Bean
-    @ConditionalOnMissingBean(name = "swaggerSpringfoxDocket")
-    public Docket swaggerSpringfoxDocket(SwaggerProperties swaggerProperties) {
+    @ConditionalOnMissingBean(name = "valtimoOpenAPI")
+    public OpenAPI valtimoOpenAPI(SwaggerProperties swaggerProperties) {
```
Rick-Ritense, line 54: Shouldn't the SwaggerProperties class be renamed as well?

web/src/main/java/com/ritense/valtimo/web/autoconfigure/SwaggerAutoConfiguration.java
```diff
@@ -45,10 +39,8 @@
 @Configuration
 @EnableConfigurationProperties(value = {SwaggerProperties.class})
 @ConditionalOnProperty(value = {"valtimo.swagger.enabled"})
-@EnableSwagger2
```
Rick-Ritense, line 48: Since this class no longer uses swagger I think this class and the security configurer should be renamed as well

Same with the log messages below

web/src/main/java/com/ritense/valtimo/web/autoconfigure/SwaggerAutoConfiguration.java
ivo-ritense: I suppose while i'm working on this i could rename all swagger related things to OpenAPI.

Rick-Ritense (approved):

marijnritense (approved):

PR #373, opened 2022-11-02
fix: Shouln't create search field with same key

Klaas-Ritense:


ivo-ritense:
We should also have a test to cover this.

Rick-Ritense (approved):

PR #377, opened 2022-11-05
Improve error handling and logging for unsupported JWT tokens.

erik-ritense:


erik-ritense:
This PR is inspired by the situation when there is a valid JWT without an email address. In the current situation this results in two NullPointerException stacktraces (one for email and one for authentication).

keycloak-iam/src/main/java/com/valtimo/keycloak/security/jwt/authentication/KeycloakTokenAuthenticator.java
```diff
@@ -50,19 +50,19 @@ public KeycloakTokenAuthenticator(String keycloakClient) {
     public boolean supports(Claims claims) {
         try {
             final String email = getEmail(claims);
-            if (email.isBlank()) {
-                logger.debug("Support failed: email is blank");
+            if (email == null || email.isBlank()) {
```
marijnritense, line 53: I think this brings up an interesting point. If the value is there but empty, you can argue it's deliberate (and so we can say the email should not be blank), but if the field is not there altogether, then email is not left blank, because it isn't present. Maybe this warrants a separate if statement instead, stating "email must be present"?

marijnritense:
@erik-ritense I would like to see a documentation PR as well before approving this.

keycloak-iam/src/main/java/com/valtimo/keycloak/security/jwt/authentication/KeycloakTokenAuthenticator.java
erik-ritense: Agreed, I can make the null check a separate if statement

Klaas-Ritense (approved):

core/src/main/java/com/ritense/valtimo/security/jwt/exception/TokenAuthenticatorNotFoundException.java
```diff
@@ -0,0 +1,8 @@
+package com.ritense.valtimo.security.jwt.exception;
```
Klaas-Ritense, line 1: The build pipeline is failing because this new class is missing the Copyright information
```
[ant:checkstyle] [ERROR] /tmp/jenkins-377d2bce/workspace/ltimo_-_Backend_libraries_PR-377/core/src/main/java/com/ritense/valtimo/security/jwt/exception/TokenAuthenticatorNotFoundException.java:1: Missing a header - not enough lines in file. [RegexpHeader]
```

erik-ritense: Thanks, I was not aware of this . I will add the copyright header.

PR #383, opened 2022-11-10
#37574 extend document search possibilities

Klaas-Ritense:


document/src/main/java/com/ritense/document/domain/search/SearchRequestValue.java
```diff
@@ -0,0 +1,103 @@
...
+import java.util.List;
+import java.util.stream.Collectors;
+
+import static com.ritense.valtimo.contract.utils.AssertionConcern.assertArgumentFalse;
+import static com.ritense.valtimo.contract.utils.AssertionConcern.assertArgumentTrue;
+
+/**
+ *
+ */
+public class SearchRequestValue {
```
Rick-Ritense, line 32: Empty comment?

document/src/test/resources/config/application.yml
```diff
@@ -46,3 +46,8 @@ valtimo:
     versioning.enabled: true
     plugin:
         encryption-secret: "abcdefghijklmnop"
+
+logging:
+    level:
+        org.hibernate.SQL: DEBUG
+        org.hibernate.type: TRACE
```
Rick-Ritense, line 53: Do we want to keep this?
Seems obsessive for non-development purposes

Klaas-Ritense: This is used in the integration-tests and is very useful for debugging issues with the document search

contract/src/main/java/com/ritense/valtimo/contract/database/QueryDialectHelper.java
```diff
@@ -23,7 +23,9 @@
 
 public interface QueryDialectHelper {
 
-    Expression<?> getJsonValueExpression(CriteriaBuilder cb, Path column, String path);
+    Expression<String> getJsonValueExpression(CriteriaBuilder cb, Path column, String path);
```
ivo-ritense, line 26: Is there a reason why this is 2 different methods? Can't we just pass in the string class for the other one and get rid of this method?

document/src/main/java/com/ritense/document/domain/search/SearchRequest2.java
```diff
@@ -0,0 +1,154 @@
...
+ */
+
+package com.ritense.document.domain.search;
+
+import java.util.ArrayList;
+import java.util.List;
+import java.util.stream.Collectors;
+
+public class SearchRequest2 {
+    private SearchOperator searchOperator = SearchOperator.OR;
```
ivo-ritense, line 24: i would say the default should be AND

document/src/main/java/com/ritense/document/service/impl/JsonSchemaDocumentSearchService.java
```diff
@@ -158,44 +244,238 @@ private void addJsonFieldPredicates(CriteriaBuilder cb, Root<JsonSchemaDocument>
...
+        } else {
+            throw new IllegalArgumentException("Search path doesn't start with known prefix: '" + searchCriteria.getPath() + "'");
+        }
+
+        var rangeFrom = searchCriteria.<T>getRangeFrom();
+        var rangeTo = searchCriteria.<T>getRangeTo();
+
+        switch (searchCriteria.getSearchType()) {
+            case LIKE:
+                return cb.or(searchLike(cb, value, searchCriteria.getValues()));
```
ivo-ritense, line 306: Why is the `cb.or` here? Do that still work with an and search?

ivo-ritense: I guess it's to support lists of strings?

Klaas-Ritense: Yes, its to support this:
```
new SearchRequest2().addOtherFilters(new SearchRequest2.SearchCriteria2()
    .path("doc:customer.fullName")
    .searchType(LIKE)
    .addValue("John D")
    .addValue("Jane D"))
```

contract/src/main/java/com/ritense/valtimo/contract/database/QueryDialectHelper.java
Klaas-Ritense: Yes, the method is no longer needed. But there is one project that uses this method still. So removing it is a breaking change

document/src/main/java/com/ritense/document/domain/search/SearchRequest2.java
Klaas-Ritense: Changed

PR #384, opened 2022-11-10
Improved performance of document definition list endpoint.

ivo-ritense:


Klaas-Ritense (approved):

document/src/main/java/com/ritense/document/repository/impl/JsonSchemaDocumentDefinitionRepository.java
```diff
@@ -27,24 +27,55 @@
...
-    @Query(""
-        + "SELECT  distinct dd "
-        + "FROM    JsonSchemaDocumentDefinition dd "
-        + "INNER JOIN JsonSchemaDocumentDefinitionRole ddRole ON ddRole.id.documentDefinitionName = dd.id.name AND ddRole.id.role in :roles "
-        + "WHERE   dd.id.version = (" +
-        "   SELECT max(dd2.id.version) " +
-        "   FROM JsonSchemaDocumentDefinition dd2 " +
-        "   WHERE dd2.id.name = dd.id.name " +
-        ") ")
+    @Query(nativeQuery = true, value = "" +
```
Tom-Ritense, line 30: I want to let u know that native queries have a downside. Using Hibernate filters is not possible in conjunction with native queries. I say this because filters can be tool for a multi-tenacy approach of filtering data.

marijnritense: Good point. In this case we've observed an improvement on performance of up to 50 times, reducing a call from 10 seconds to less than half a second. But what you're saying is definitely something to keep in mind.

Tom-Ritense: Note that staying in HSQL syntax you could improve the maybe also.

ivo-ritense: As discussed: The reason we do this as a native query is because the where in has bad performance in MySQL. The fix for that is the inner join that you see in this native query. The same is not possible in HQL, that's why we opted for the native query. For now we're merging this in the understanding that we might have to make manual changes if we introduce hibernate filters in the future.

PR #402, opened 2022-11-22
#42785 [BE] Resolve value of all process link properties

Klaas-Ritense:


document-generation/smartdocuments/src/test/kotlin/com/ritense/smartdocuments/plugin/SmartDocumentsPluginIntegrationTest.kt
```diff
@@ -170,6 +161,55 @@ class SmartDocumentsPluginIntegrationTest : BaseSmartDocumentsIntegrationTest()
         )
     }
 
+    @Test
+    fun `should resolve template-name value from process-variable`() {
```
marijnritense, line 165: Do we want to include a test to validate how Valtimo responds when the process variable can't be found?

marijnritense (approved):

PR #412, opened 2022-11-29
Override existing Camunda endpoint

luis-soares-ritense:


core/src/test/resources/examples/bpmn/shouldDeploy.xml
```diff
@@ -0,0 +1,46 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
Klaas-Ritense, line 1: `shouldDeploy.xml` I guess the .xml is to prevent Camunda from autodeploying the BPMN?

Klaas-Ritense (approved):

core/src/test/resources/examples/bpmn/shouldDeploy.xml
luis-soares-ritense: yes it is. shouldn't it be looking only to the bpmn folder?

Klaas-Ritense: Camunda does the deployment for us. We haven't really had a reason to change it. See:
https://github.com/camunda/camunda-bpm-platform/blob/master/spring-boot-starter/starter/src/main/java/org/camunda/bpm/spring/boot/starter/property/CamundaBpmProperties.java#L41

luis-soares-ritense: https://docs.camunda.org/manual/7.8/user-guide/spring-boot-integration/configuration/#camunda-engine-properties

we could add  camunda.bpm.deployment-resource-pattern: classpath:bpmn/*.bpmn

PR #425, opened 2022-12-09
Bugfix: Search on date

Klaas-Ritense:
Searching on a date gives an error when the date is saved as an empty string in the document JSON.

The error Postgres SQL look a little like this:
```
SELECT jsonb_path_query_first('{"myDate": ""}','$.myDate')::TEXT::DATE
```
Gives error:
```
SQL Error [22007]: ERROR: invalid input syntax for type date: """"
```

Solution:
```
SELECT CASE WHEN(jsonb_path_query_first('{"myDate": ""}','$.myDate')::TEXT) IN('""') THEN null ELSE BTRIM(jsonb_path_query_first('{"myDate": ""}','$.myDate')::TEXT, '"')::DATE END
```

Its beautiful


theo-ritense (approved):

theo-ritense:
Beautiful

PR #426, opened 2022-12-09
#42594: Prevent creation of duplicate form flow instances

Klaas-Ritense:
Clicking twice to open a user task, sometimes creates two FormFlowInstances. Which will result in an error


theo-ritense (approved):
It doesn't harm to add @Transactional here, but does it really fix the issue? Really depends on the type of locking I guess.

PR #428, opened 2022-12-12
#42596 FormFlowInstanceId available in step

Klaas-Ritense:


form-flow/src/main/kotlin/com/ritense/formflow/domain/instance/FormFlowStepInstance.kt
```diff
@@ -79,6 +79,9 @@ data class FormFlowStepInstance(
                 "key" to stepKey,
                 "submissionData" to instance.getSubmissionDataContext()
             ),
+            "instance" to mapOf(
```
marijnritense, line 82: Maybe just a variable at the lowest level, so instanceId? Additional properties are also on this same level.

ivo-ritense (approved):

PR #431, opened 2022-12-13
#42597 Conditional Form Flow steps

Klaas-Ritense:


form-flow/build.gradle
```diff
@@ -27,6 +27,7 @@ dependencies {
     implementation "org.springframework.boot:spring-boot-starter-data-jpa"
     implementation "org.springframework.boot:spring-boot-starter"
     implementation "org.springframework.boot:spring-boot-starter-validation"
+    implementation "org.springframework.integration:spring-integration-core:5.5.10"
```
ivo-ritense, line 30: This is another reason we need to merge marjo10. we don't want versions like this in modules again, but in the properties file. Can you rebase this branch on major 10 and merge it to there?

Klaas-Ritense: Changed

luis-soares-ritense (approved):

PR #432, opened 2022-12-14
#no-tp: Note API for update and delete

Klaas-Ritense:


notes/src/main/kotlin/com/ritense/note/service/NoteService.kt
```diff
@@ -53,6 +57,35 @@ class NoteService(
         return node
     }
 
+    fun editNote(noteId: UUID, noteContent: String): Note {
+        logger.debug { "Update note with id '$noteId'" }
+        SecurityUtils.getCurrentUserLogin()
+        val note = getNoteById(noteId)
+        verifyCurrentUserHasAccessToNote(note)
+        note.content = noteContent
```
theo-ritense, line 65: instead of having content be a `var`, can't we make Note a data class and use .copy() ?

Klaas-Ritense: Changed it. Seems to work well

luis-soares-ritense (approved):

marijnritense (approved):

PR #433, opened 2022-12-15
#45858 form flow store submission data when moving to the previous step

Klaas-Ritense:


marijnritense:
I don't think this covers the story entirely, as it also describes what should happen when closing the modal in FE. Is the intention here to deal with these two cases separately?

form-flow/src/main/kotlin/com/ritense/formflow/domain/instance/FormFlowStepInstance.kt
```diff
@@ -52,7 +52,11 @@ data class FormFlowStepInstance(
     val definition: FormFlowStep
         get() = instance.formFlowDefinition.getStepByKey(stepKey)
 
-    fun back() {
+    fun back(incompleteSubmissionData: String? = null) {
```
Klaas-Ritense, line 55: @marijnritense When the user in the browser click on back, the user will get three options:
- Click 'Yes, save data'. The FE should call `/back` _and_ send the incompleteSubmissionData. 
- Click 'No, don't save data'. The FE should call `/back` but _not_ send any data. The mothod above, will receive `null`
- Click 'Cancel'. The FE closes the popup but shouldn't call `/back`

marijnritense: image

If you click outside of the form flow, you should get those three options as well, but then it shouldn't call /back as we don't want to go back a step, we just want to be able to continue where we left off next time we open the task.

PR #446, opened 2022-12-20
created the object management module

Niek-Ritense:


zgw/object-management/docker-compose-override.yml
```diff
@@ -0,0 +1,7 @@
+version: '3.9'
+services:
+    db:
+        ports:
+            - "3330:5432"
```
Klaas-Ritense, line 5: port `3330` is already taken. See: 'docker-compose-base-test.yml'

settings.gradle
```diff
@@ -65,6 +65,7 @@ include(
     "zgw:catalogi-api",
     "zgw:documenten-api",
     "zgw:notificaties-api",
+    "zgw:object-management",
```
Klaas-Ritense, line 68: This new module should also be added as a dependency to `/app/gzac/build.gradle`

zgw/object-management/src/main/resources/META-INF/spring.factories
```diff
@@ -0,0 +1,2 @@
+org.springframework.boot.autoconfigure.EnableAutoConfiguration=\
+  com.ritense.documentenapi.DocumentenApiAutoConfiguration
```
Klaas-Ritense, line 2: This autoconfiguration is not correct. This new module doesnt have a autoconfiguration yet. Either:
- create empty autocofiguration in this module. or
- make this file empty for now

Klaas-Ritense (approved):

PR #466, opened 2023-01-09
Bugfix/fix aws inspector findings for major 9

theo-ritense:


gradle.properties
```diff
@@ -16,16 +16,16 @@
 projectGroup=com.ritense.valtimo
 projectName=valtimo-backend-libraries
 kotlinVersion=1.6.10
-springBootVersion=2.5.12
-springDependencyManagementVersion=1.0.11.RELEASE
+springBootVersion=2.5.14
```
Klaas-Ritense, line 19: Valtimo has issues when using Spring 2.5.14. See: https://ritensebv.slack.com/archives/C1ZHMQGF7/p1671638923456599?thread_ts=1671636650.325239&cid=C1ZHMQGF7

theo-ritense: As discussed, we were unable to solve the security issues we had by staying on 2.5.x. We've upgraded to Spring Boot 2.7.7 because it's more important to have a secure application then a compatible one.

Please check again.

PR #467, opened 2023-01-09
Load list column definitions on startup

ivo-ritense:


case/src/main/kotlin/com/ritense/case/service/CaseListDeploymentService.kt
```diff
@@ -0,0 +1,98 @@
+/*
+ * Copyright 2015-2022 Ritense BV, the Netherlands.
```
marijnritense, line 2: 2023?

case/src/main/kotlin/com/ritense/case/service/CaseListDeploymentService.kt
```diff
@@ -0,0 +1,98 @@
...
+        return resourcePatternResolver.getResources(CASE_LIST_DEFINITIONS_PATH)
+    }
+
+    private fun loadCaseListSchemaResource(): Resource {
+        return resourcePatternResolver.getResource(CASE_LIST_SCHEMA_PATH)
+    }
+
+    companion object {
+        internal const val CASE_LIST_SCHEMA_PATH = "classpath:config/case/schema/case-list.schema.json"
+        internal const val CASE_LIST_DEFINITIONS_PATH = "classpath:config/case/list/*.json"
```
marijnritense, line 95: I forget, didn't we talk about wanting to allow implementation to separate case definitions as well?

marijnritense (approved):

case/src/main/resources/config/case/schema/case-list.schema.json
```diff
@@ -0,0 +1,51 @@
+{
+    "$schema": "http://json-schema.org/draft-07/schema#",
+    "$id": "http://json-schema.org/draft-07/schema#",
+    "title": "Form flow definition",
+    "description": "Representation of a form flow definition",
```
ThomasRitense, line 5: Representation of form flow?

luis-soares-ritense: fixed in new commit

luis-soares-ritense (approved):

PR #478, opened 2023-01-11
feat: 47087 interfacing with Notificaties API

anna-ritense:
* adds liquibase configuration for creating new `notificaties_api_abonnement_link` table so that created subsciptions can be stored in the database
* adds client for communicating with Notificaties Api. The client implements createKanaal, getKanalen, createAbonnement and deleteAbonnement methods
* adds methods to the plugin for creating and deleting subscriptions so that other plugins can make use of Notificaties API

### DIFF
--- a/zgw/notificaties-api/build.gradle
+++ b/zgw/notificaties-api/build.gradle
@@ -36,7 +36,6 @@ dependencies {
     //test
     testImplementation "org.jetbrains.kotlin:kotlin-test"
     testImplementation "org.junit.jupiter:junit-jupiter"
-    testImplementation "org.springframework.boot:spring-boot-starter-webflux"
     testImplementation "com.squareup.okhttp3:mockwebserver:${okhttpVersion}"
     testImplementation "com.squareup.okhttp3:okhttp:${okhttpVersion}"
     testImplementation "org.mockito.kotlin:mockito-kotlin:${mockitoKotlinVersion}"
--- a/zgw/notificaties-api/build.gradle
+++ b/zgw/notificaties-api/build.gradle
@@ -22,22 +22,28 @@ dockerCompose {
 }
 
 dependencies {
+    implementation project(':contract')
     implementation project(':plugin')
     implementation project(':openzaak')
 
-    implementation "io.github.microutils:kotlin-logging:${kotlinLoggingVersion}"
-    implementation "org.springframework.boot:spring-boot-starter-webflux"
     implementation "org.springframework.boot:spring-boot-starter-security"
+    implementation "org.springframework.boot:spring-boot-starter-data-jpa"
+    implementation "org.springframework.boot:spring-boot-starter-webflux"
+    implementation "io.github.microutils:kotlin-logging:${kotlinLoggingVersion}"
+    implementation "org.jetbrains.kotlinx:kotlinx-coroutines-core:$kotlinCoroutinesVersion"
+    implementation "org.jetbrains.kotlinx:kotlinx-coroutines-reactor:$kotlinCoroutinesVersion"
 
-    testImplementation project(":contract")
-
+    //test
     testImplementation "org.jetbrains.kotlin:kotlin-test"
     testImplementation "org.junit.jupiter:junit-jupiter"
-    testImplementation "org.springframework.boot:spring-boot-starter-test"
+    testImplementation "org.springframework.boot:spring-boot-starter-webflux"
     testImplementation "com.squareup.okhttp3:mockwebserver:${okhttpVersion}"
     testImplementation "com.squareup.okhttp3:okhttp:${okhttpVersion}"
-    //Kotlin
     testImplementation "org.mockito.kotlin:mockito-kotlin:${mockitoKotlinVersion}"
+    testImplementation "org.jetbrains.kotlinx:kotlinx-coroutines-core:$kotlinCoroutinesVersion"
+    testImplementation "org.jetbrains.kotlinx:kotlinx-coroutines-test:$kotlinCoroutinesVersion"
+
+
 
     jar {
         enabled = true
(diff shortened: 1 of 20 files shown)


zgw/notificaties-api/build.gradle
```diff
@@ -22,22 +22,28 @@ dockerCompose {
...
+    implementation "org.jetbrains.kotlinx:kotlinx-coroutines-core:$kotlinCoroutinesVersion"
+    implementation "org.jetbrains.kotlinx:kotlinx-coroutines-reactor:$kotlinCoroutinesVersion"
 
-    testImplementation project(":contract")
-
+    //test
     testImplementation "org.jetbrains.kotlin:kotlin-test"
     testImplementation "org.junit.jupiter:junit-jupiter"
-    testImplementation "org.springframework.boot:spring-boot-starter-test"
+    testImplementation "org.springframework.boot:spring-boot-starter-webflux"
```
Klaas-Ritense, line 39: This dependency is also defined above

zgw/notificaties-api/src/main/kotlin/com/ritense/notificatiesapi/autoconfigure/NotificatiesApiLiquibaseAutoConfiguration.kt
```diff
@@ -0,0 +1,36 @@
...
+import org.springframework.context.annotation.Bean
+import org.springframework.context.annotation.Configuration
+import org.springframework.core.Ordered
+import org.springframework.core.annotation.Order
+
+@Configuration
+@ConditionalOnClass(DataSource::class)
+class NotificatiesApiLiquibaseAutoConfiguration {
+
+    @Order(Ordered.HIGHEST_PRECEDENCE + 12)
```
Klaas-Ritense, line 31: Could you change this to `+ 25`?
To ensure that this liquibase script is executed after all other liquibase scripts inside Valtimo

zgw/notificaties-api/src/main/kotlin/com/ritense/notificatiesapi/repository/AbonnementLinkRepository.kt
```diff
@@ -0,0 +1,23 @@
...
+ * limitations under the License.
+ */
+
+package com.ritense.notificatiesapi.repository
+
+import com.ritense.notificatiesapi.domain.NotificatiesApiAbonnementLink
+import com.ritense.notificatiesapi.domain.NotificatiesApiConfigurationId
+import org.springframework.data.jpa.repository.JpaRepository
+
+interface AbonnementLinkRepository : JpaRepository<NotificatiesApiAbonnementLink, NotificatiesApiConfigurationId>
```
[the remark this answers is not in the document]
anna-ritense: Yea, I noticed this earlier today and already made the change before this comment, just hadn't pushed yet. Thanks for noticing!

zgw/notificaties-api/build.gradle
anna-ritense: Thanks, didn't notice the duplicate.

zgw/notificaties-api/src/main/kotlin/com/ritense/notificatiesapi/autoconfigure/NotificatiesApiLiquibaseAutoConfiguration.kt
anna-ritense: Adressed.

Is this specified somewhere what order we want what liquibase changesets to run? I assume the reason here is so that valtimo/gzac core can always ensure they can run their changesets before any additional features are deployed.
Maybe worth noting this somewhere, so that people who want to contribute know that precedence <25 is reserved.

zgw/notificaties-api/src/main/kotlin/com/ritense/notificatiesapi/NotificatiesApiPlugin.kt
```diff
@@ -16,19 +16,105 @@
...
+                )
+            )
+        }
+    }
+
+    fun deleteAbonnement(pluginConfigurationId: PluginConfigurationId) {
+        val pluginId = NotificatiesApiConfigurationId(pluginConfigurationId.id)
+
+        abonnementLinkRepository.findById(pluginId)
+            .ifPresent {
```
marijnritense, line 82: If an abonnement isn't present when this method is called, shouldn't that at least result in a log message as well?

anna-ritense: This is the way it was done in the Connector for opennotificaties. If it's not present when `deleteAbonnement` is called then it doesn't matter since the expected outcome of this method is to remove it from the database.

I do however see the need to log this, since it might point towards an issue with the database if an entity that is expected to be there isn't.
I'll add a log entry.

zgw/notificaties-api/src/main/kotlin/com/ritense/notificatiesapi/NotificatiesApiPlugin.kt
```diff
@@ -16,19 +16,105 @@
...
+    private val abonnementLinkRepository: NotificatiesApiAbonnementLinkRepository
+) {
     @PluginProperty(key = "url", secret = false)
     lateinit var url: URI
 
     @PluginProperty(key = "authenticationPluginConfiguration", secret = false)
     lateinit var authenticationPluginConfiguration: NotificatiesApiAuthentication
+
+    fun createAbonnement(
+        pluginConfigurationId: PluginConfigurationId,
```
marijnritense, line 51: Since we're already within the context of a plugin instance, maybe it makes more sense to provide the plugin configuration id another way, e.g. through the pluginfactory.

anna-ritense: Also fair thing to consider. I do wonder how much that adds to this specific plugin, given that it's not meant to do much on it's own and is mainly a plugin that other plugins/services depend on.

Thoughts? @Klaas-Ritense

Klaas-Ritense (approved):

PR #479, opened 2023-01-11
htmlDependencyReport for all submodules

rensiritense:
Change to make htmlDependencyReport also report on all submodules


Hildebrand-Ritense (changes requested):
ok, I think it's best to make a small change here; to move the htmlDependencyReport configuration higher up in the file (e.g. before / after sonarqube config (lines 48-53)). This because the runtime insertion of the publishing gradle configuration is based on linecount - X lines. 

Having the config block at the current position makes it so that you'd have to update that while if you put it in the before mentioned place you won't.

rensiritense:
Moved to line 48

Hildebrand-Ritense (approved):
lgtm

PR #492, opened 2023-01-17
#47145 Zaken Api plugin - new plugin action 'Create zaak'

Klaas-Ritense:


openzaak/src/main/kotlin/com/ritense/openzaak/domain/mapping/impl/ZaakInstanceLinks.kt
```diff
@@ -16,7 +16,9 @@
 
 package com.ritense.openzaak.domain.mapping.impl
 
+import com.ritense.zakenapi.domain.ZaakInstanceLink
+
 class ZaakInstanceLinks : HashSet<ZaakInstanceLink> {
```
luis-soares-ritense, line 21: is this really necessary?

Klaas-Ritense: yes, 
```
import com.ritense.zakenapi.domain.ZaakInstanceLink
```
is needed because it is moved to a different module

luis-soares-ritense: i mean do we need an object that is just an HashSet of ZaakInstanceLink

Klaas-Ritense: Just checked it. This class is really old and isn't used anywhere. So i removed it

zgw/zaken-api/src/main/kotlin/com/ritense/zakenapi/repository/ZaakInstanceLinkRepository.kt
```diff
@@ -0,0 +1,32 @@
...
+import com.ritense.zakenapi.domain.ZaakInstanceLinkId
+import org.springframework.data.jpa.repository.JpaRepository
+import org.springframework.stereotype.Repository
+import java.net.URI
+import java.util.UUID
+
+@Repository
+interface ZaakInstanceLinkRepository : JpaRepository<ZaakInstanceLink, ZaakInstanceLinkId> {
+
+    fun findByDocumentId(documentId: UUID): ZaakInstanceLink?
```
marijnritense, line 29: I wonder if it will always be true there is a 1:1 relation between a zaak and a document.

marijnritense (approved):

zgw/zaken-api/src/main/kotlin/com/ritense/zakenapi/repository/ZaakInstanceLinkRepository.kt
Klaas-Ritense: For every case in Valtimo there should only ever be one Zaak in OpenZaak

PR #493, opened 2023-01-18
Feature/bill of materials

marijnritense:
Also depends on an internal PR. I tried to stick to the core modules of each edition (Core and GZAC), but a lot of dependencies in these modules are also transitive, and as a result will also end up in implementations who want to use Valtimo. This takes care of a lot of the dependency management as well, which is why it is so slim. I considered removing Spring from here, but I think setting the version in here makes sense.


theo-ritense (commented):
I still feel like this shouldn't be called a BOM, because it isn't. Can we please rename the modules to valtimo-dependencies and valtimo-gzac-dependencies?

theo-ritense (approved):

PR #503, opened 2023-01-25
feat: 47126 plugin events

anna-ritense:
* adds the ability to run specific methods of a plugin instance when it's corresponding Plugin Configuration is created, updated or deleted.
* fixes various errors in existing Integration and Unit Tests (marijnritense)
* fixes bug with NotificatiesApiAbonnementLink having the wrong type where an Abonnement could not be found in the repository because of a type mismatch on the id column
* changes column name in table notificaties_api_abonnement_link to appropriately represent the value it holds (*Breaking Change*)
* changes methods of Notificaties Api plugin to not require any arguments and use class members instead (*Breaking Change*)
* fixes bug in NotificatiesApiClient where the response from OpenNotificaties could not be mapped to an Abonnement DTO, because of a missing auth property in the response body


plugin/src/main/kotlin/com/ritense/plugin/service/PluginService.kt
```diff
@@ -85,9 +88,19 @@ class PluginService(
...
-        return pluginConfigurationRepository.save(
+        val pluginConfiguration = pluginConfigurationRepository.save(
             PluginConfiguration(PluginConfigurationId.newId(), title, properties, pluginDefinition)
         )
+
+        try {
+            pluginConfiguration.runAllPluginEvents(PluginEvent.EventType.CREATE)
+
+        } catch (e: Exception) {
+            pluginConfigurationRepository.deleteById(pluginConfiguration.id)
```
Klaas-Ritense, line 99: The method 'createPluginConfiguration' can be `@Transactional`. 

That way the pluginConfiguration won't have to be deleted here. Because the exception will trigger a rollback which will rollback the `.save()`

plugin/src/main/kotlin/com/ritense/plugin/service/PluginService.kt
```diff
@@ -307,14 +350,34 @@ class PluginService(
...
+        filter: (JsonNode) -> Boolean
+    ): PluginConfiguration? {
         val configurations = pluginConfigurationRepository.findByPluginDefinitionKey(pluginDefinitionKey)
         return configurations.firstOrNull { config ->
             val configProperties = config.properties
             configProperties != null && filter(configProperties)
         }
     }
 
+    private fun PluginConfiguration.runAllPluginEvents(eventType: PluginEvent.EventType): PluginConfiguration {
```
Klaas-Ritense, line 364: The overload function can just be a service method

anna-ritense: Yes, it can be a normal method, but that removes the ability to chain it. Having it be an extension method for the PluginConfiguration class is purely a qol choice.
That said, if you are against this, then I will rework it.

plugin/src/main/kotlin/com/ritense/plugin/service/PluginService.kt
anna-ritense: Yes, we tried this and discussed this with Marijn when fixing existing Integratin and Unit Tests.
Let's discuss this.

zgw/notificaties-api/src/main/kotlin/com/ritense/notificatiesapi/client/NotificatiesApiClient.kt
```diff
@@ -58,11 +58,16 @@ class NotificatiesApiClient(
...
+    internal suspend fun createKanaal(
+        authentication: NotificatiesApiAuthentication,
+        baseUrl: URI,
+        kanaal: Kanaal
+    ): Kanaal {
         return buildNotificatiesWebClient(authentication, baseUrl)
             .post()
             .uri("kanaal")
             .contentType(MediaType.APPLICATION_JSON)
+            .headers { it.setAll(mapOf("header1" to "value1", "header2" to "value2")) }
```
Klaas-Ritense, line 70: This shouldn't be there

anna-ritense: whoops, this was a mistake and was used to make a point earlier haha, sorry! 🙏

Klaas-Ritense (approved):

PR #509, opened 2023-01-27
Bugfix/fixes

Klaas-Ritense:
This PR includes:
- Zaken API plugin default configuration for local development
- Rename property 'systemProcessDefinitionKey' to 'processToStart' after request from Thomas
- Will now crash the application on startup when a document-definition-schema contains an error
- Delete duplicate form IO definitions
- Fix milestones module dependency
- Sort the list of plugin definitions and configurations in browser
- NotificatieAPI url remove the '/api/v1' part
- Objecten API client, remove the '/api/v1' part


zgw/notificaties-api/src/main/kotlin/com/ritense/notificatiesapi/client/NotificatiesApiClient.kt
```diff
@@ -38,7 +38,7 @@ class NotificatiesApiClient(
 
         return buildNotificatiesWebClient(authentication, baseUrl)
             .post()
-            .uri("/api/v1/abonnement")
+            .uri("abonnement")
```
Klaas-Ritense, line 41: @anna-ritense Fyi

anna-ritense: I want to discuss this.

marijnritense: @Klaas-Ritense I don't think this change should go through. The code is meant to work for /api/v1/, so it should be a part of this line here. It's possible when a v2 of the API will come out, it will not work with the existing code. This class is the client for v1, right?

marijnritense (approved):

PR #517, opened 2023-01-30
SSE implementation for Camunda task events (generic base events)

Tom-Ritense:


core/src/main/java/com/ritense/valtimo/sse/domain/listener/TaskUpdateListener.kt
```diff
@@ -0,0 +1,41 @@
...
+import com.ritense.valtimo.sse.event.TaskUpdateSseEvent
+import com.ritense.valtimo.sse.service.SseSubscriptionService
+import org.camunda.bpm.engine.delegate.DelegateTask
+import org.springframework.transaction.event.TransactionalEventListener
+
+class TaskUpdateListener(
+    private val sseSubscriptionService: SseSubscriptionService
+) {
+
+    @TransactionalEventListener(
```
Klaas-Ritense, line 28: The `@TransactionalEventListener` doesn't always catch the event:

> If the event is not published within an active transaction, the event is discarded unless the fallbackExecution flag is explicitly set.

WouterG:
Paired with frontend PR https://github.com/valtimo-platform/valtimo-frontend-libraries/pull/383

Frontend receives the events, does not act on them, in our implementation we do, we could also port that over to the frontend PR.
Generally the idea of the SSE service in backend and frontend is for implementations to allow them to plug in events and listen to them for custom things, but default events can be added too.

core/src/test/kotlin/com/ritense/valtimo/sse/service/SseSubscriptionServiceTest.kt
```diff
@@ -0,0 +1,48 @@
+package com.ritense.valtimo.sse.service
```
Klaas-Ritense, line 1: Missing copyright header

Klaas-Ritense (approved):

PR #522, opened 2023-01-31
Fix query

Klaas-Ritense:


form/src/main/resources/config/liquibase/changelog/20220123-add-name-unique-constraint-changelog.xml
```diff
@@ -42,9 +42,11 @@
             DELETE
             FROM form_io_form_definition
             WHERE id NOT IN (
-                SELECT MIN(id)
-                FROM form_io_form_definition
-                GROUP BY name
+                SELECT fid FROM (
```
marijnritense, line 45: Doesn't this change the checksum?

Klaas-Ritense: It does. However, no one is able to execute this query because its broken. 

context:
https://stackoverflow.com/questions/45494/mysql-error-1093-cant-specify-target-table-for-update-in-from-clause#answer-9843719

ivo-ritense (approved):

PR #525, opened 2023-01-31
Feature/47061 Notificaties API - receiving notifications

luis-soares-ritense:


zgw/notificaties-api/src/main/kotlin/com/ritense/notificatiesapi/autoconfigure/NotificatiesApiAutoConfiguration.kt
```diff
@@ -56,4 +61,37 @@ class NotificatiesApiAutoConfiguration {
             abonnementLinkRepository
         )
     }
+
+    @Bean
+    @ConditionalOnMissingBean(WebClient::class)
+    fun notificatiesApiWebClient(): WebClient {
```
Klaas-Ritense, line 67: We stopped creating webclient beans since yesterday. See: [this PR](https://github.com/valtimo-platform/valtimo-backend-libraries/pull/520/files#diff-3f9d5687e61501ceebbdfc620222693f7ef0b05334d458ca8ddf6a81adeaaf74L50-L71)

zgw/notificaties-api/src/main/kotlin/com/ritense/notificatiesapi/security/config/NotificatiesApiHttpSecurityConfigurer.kt
```diff
@@ -0,0 +1,34 @@
...
+import com.ritense.valtimo.contract.security.config.HttpConfigurerConfigurationException
+import com.ritense.valtimo.contract.security.config.HttpSecurityConfigurer
+import org.springframework.http.HttpMethod
+import org.springframework.security.config.annotation.web.builders.HttpSecurity
+
+class NotificatiesApiHttpSecurityConfigurer : HttpSecurityConfigurer {
+    override fun configure(http: HttpSecurity) {
+        try {
+            http.authorizeRequests()
+                .antMatchers(HttpMethod.POST, "/api/v1/notificaties/send")
```
Klaas-Ritense, line 28: Not sure about the `/send`. Since it doesn't really send anything. It only receives notifications

zgw/notificaties-api/src/main/kotlin/com/ritense/notificatiesapi/web/rest/NotificatiesApiResource.kt
```diff
@@ -0,0 +1,52 @@
...
+        @RequestBody notification: NotificatiesApiNotificationReceivedEvent,
+        @RequestHeader("Authorization") authHeader: String?
+    ): ResponseEntity<Void> {
+        return if (authHeader != null) {
+            try {
+                notificatiesApiService.findAbonnementSubscription(authHeader)
+                notificatiesApiService.handle(notification)
+                ResponseEntity.noContent().build()
+            } catch (ex: RuntimeException) {
+                ResponseEntity.status(HttpStatus.UNAUTHORIZED).build()
```
Klaas-Ritense, line 45: A RuntimeException is should not result in an UNAUTHORIZED message. Because the notification can be handled in may different listeners that can all throw an exception.

Exception handeling is e big reason of why we are rewriting the existing connector to a new plugin. In the previous connector, exceptions were often not visible.

Also the exception message is being swallowed here. The exception message with stacktrace should be returned by this endpoint. The exception will then end up in the logs of OpenNotificaties. This helps with debugging

zgw/notificaties-api/src/main/kotlin/com/ritense/notificatiesapi/autoconfigure/NotificatiesApiAutoConfiguration.kt
luis-soares-ritense: That was already there..what should be done then?

zgw/verzoek/build.gradle
```diff
@@ -27,12 +27,14 @@ dependencies {
     implementation project(":process-document")
     implementation project(":plugin")
     implementation project(":plugin-authentication:notificaties-api-authentication")
+    implementation project(":openzaak")
```
Klaas-Ritense, line 30: Can this dependency be removed?

zgw/notificaties-api/src/main/kotlin/com/ritense/notificatiesapi/security/config/NotificatiesApiHttpSecurityConfigurer.kt
Klaas-Ritense: The `/send` is still there

zgw/notificaties-api/src/main/kotlin/com/ritense/notificatiesapi/autoconfigure/NotificatiesApiAutoConfiguration.kt
Klaas-Ritense: This webclient bean should be removed

zgw/notificaties-api/src/main/kotlin/com/ritense/notificatiesapi/security/config/NotificatiesApiHttpSecurityConfigurer.kt
anna-ritense: while testing locally, I made a temporary one with this path: `/api/v1/notificatiesapi/notification`. it made sense to me that we keep the path consistent with the plugin name. would this be an option?

luis-soares-ritense: @Klaas-Ritense it can't be i removed it.
@anna-ritense can we compromise on "/api/v1/notificaties/notification"? my reasoning being that "notificatiesapi" is not a resource that's being targeted. just "notificaties" seems more like the resource itself.

anna-ritense: I vote for the second one, that way the resource is clearly different from the old one that's provided by the connector.

Klaas-Ritense: Luis, can you also change the url in the `NotificatiesApiHttpSecurityConfigurer.kt`

zgw/verzoek/build.gradle
Klaas-Ritense: > Can this dependency be removed?

zgw/notificaties-api/src/main/kotlin/com/ritense/notificatiesapi/security/config/NotificatiesApiHttpSecurityConfigurer.kt
```diff
@@ -0,0 +1,34 @@
...
+import com.ritense.valtimo.contract.security.config.HttpConfigurerConfigurationException
+import com.ritense.valtimo.contract.security.config.HttpSecurityConfigurer
+import org.springframework.http.HttpMethod
+import org.springframework.security.config.annotation.web.builders.HttpSecurity
+
+class NotificatiesApiHttpSecurityConfigurer : HttpSecurityConfigurer {
+    override fun configure(http: HttpSecurity) {
+        try {
+            http.authorizeRequests()
+                .antMatchers(HttpMethod.POST, "api/v1/notificatiesapi/callback")
```
Klaas-Ritense, line 28:
```suggestion
                .antMatchers(HttpMethod.POST, "/api/v1/notificatiesapi/callback")
```

zgw/verzoek/build.gradle
Klaas-Ritense:
```suggestion
```

zgw/notificaties-api/src/main/kotlin/com/ritense/notificatiesapi/exception/NotificatiesException.kt
```diff
@@ -0,0 +1,22 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.ritense.notificatiesapi.exception
+
+import org.springframework.http.HttpStatus
+
+class NotificatiesException(message: String?, val status: HttpStatus) : RuntimeException(message) {
```
marijnritense, line 21: This exception doesn't really say a whole lot. Generally speaking, exceptions say something about the kind of exception that is being thrown. In this case, it mentions the domain the error is related to.

Klaas-Ritense (approved):

zgw/notificaties-api/src/main/kotlin/com/ritense/notificatiesapi/exception/NotificatiesException.kt
luis-soares-ritense: well this was only to wrap some exception that may be thrown by the listeners. each listener may have different reasons for throwing exceptions. if they want it to be captured and sent out they should extend this exception or throw it directly.

PR #532, opened 2023-02-03
added portaltaak uploaded document bpmn

luis-soares-ritense:


theo-ritense (commented):
This is not a system process. There should be a property in the bpmn flagging it as such.
Please take a look at as an example:
https://github.com/valtimo-platform/valtimo-backend-libraries/blob/main/core/src/test/resources/examples/bpmn/systemProcess.xml#L9

theo-ritense (approved):

PR #540, opened 2023-02-08
Feature/47740 complete camunda task when portal task is completed

luis-soares-ritense:


contract/src/main/java/com/ritense/valtimo/contract/json/patch/JsonPatchBuilder.java
```diff
@@ -82,8 +84,30 @@ public JsonPatchBuilder copy(JsonPointer from, JsonPointer to) {
         return this;
     }
 
+    /** Adds a JsonNode value to a json at the specified location. */
+    public JsonPatchBuilder addJsonNodeValue(JsonNode destination, JsonPointer path, JsonNode value) {
```
marijnritense, line 88: Not entirely sure if this logic is sound, at the very least it doesn't read nicely. Can you add some comments to make this easier to read?

openzaak/src/main/kotlin/com/ritense/openzaak/service/impl/ZaakInstanceLinkService.kt
```diff
@@ -26,7 +26,7 @@ import com.ritense.zakenapi.repository.ZaakInstanceLinkRepository
 import java.net.URI
 import java.util.UUID
 
-class ZaakInstanceLinkService(
+class  ZaakInstanceLinkService(
```
marijnritense, line 29: Why the extra space?

contract/src/main/java/com/ritense/valtimo/contract/json/patch/JsonPatchBuilder.java
luis-soares-ritense: well i don't really know about that stuff. i dont really know where that actually came from

zgw/portaaltaak/src/main/kotlin/com/ritense/portaaltaak/PortaalTaakEventListener.kt
```diff
@@ -0,0 +1,178 @@
...
+    private val processService: CamundaProcessService,
+    private val taskService: TaskService,
+    private val documentService: DocumentService
+) {
+
+    @EventListener(NotificatiesApiNotificationReceivedEvent::class)
+    fun processCompletePortaalTaakEvent(event: NotificatiesApiNotificationReceivedEvent) {
+        val objectType = event.kenmerken["objectType"]
+
+        if (!event.kanaal.equals("objecten", ignoreCase = true) ||
```
marijnritense, line 56: Nitpicky: it reads a bit weird to specify `ignoreCase`, but not `other`)

zgw/portaaltaak/src/main/kotlin/com/ritense/portaaltaak/PortaalTaakEventListener.kt
```diff
@@ -0,0 +1,178 @@
...
+            return
+        }
+
+        val objectManagement =
+            objectManagementService.findByObjectTypeId(objectType.substringAfterLast("/")) ?: return
+
+        // todo need clarification why this is really needed
+        pluginService.createInstance(PortaaltaakPlugin::class.java) { properties: JsonNode
+            ->
+            properties.get("objectManagementConfigurationId").textValue().equals(objectManagement.id.toString())
```
marijnritense, line 69: Shouldn't this already be a UUID? If so, a regular equals should be sufficient.

zgw/portaaltaak/src/main/kotlin/com/ritense/portaaltaak/PortaalTaakEventListener.kt
```diff
@@ -0,0 +1,178 @@
...
+            processDocumentService.getDocument(CamundaProcessInstanceId(processInstanceId), task as TaskEntity)
+        documentService.modifyDocument(
+            ModifyDocumentRequest(
+                document.id().toString(),
+                jacksonObjectMapper().valueToTree(taakObject.verzondenData),
+                document.version().toString()
+            )
+        ).also { result ->
+            if (result.errors().size > 0) {
+                throw NotificatiesException(
```
marijnritense, line 102: Still think this is a really bad exception name, as it says nothing about what the exception is, just in what domain the exception occurs.

zgw/portaaltaak/src/main/kotlin/com/ritense/portaaltaak/PortaalTaakEventListener.kt
```diff
@@ -0,0 +1,178 @@
...
+            }
+        }
+    }
+
+    internal fun getDocumentenUrls(verzondenData: JsonNode): List<String> {
+        if (verzondenData.isMissingNode || verzondenData.isNull) {
+            return emptyList()
+        }
+        if (!verzondenData.isArray) {
+            throw RuntimeException("Not an array: '/documenten'")
```
marijnritense, line 117: Maybe introduce a valtimo specific TypeMismatchException here. RuntimeException is very broad given how specific this method is.

zgw/portaaltaak/src/main/kotlin/com/ritense/portaaltaak/PortaalTaakEventListener.kt
```diff
@@ -0,0 +1,178 @@
...
+        for (documentPathNode in verzondenData) {
+            val documentUrlNode = verzondenData.at(JsonPointer.valueOf(documentPathNode.textValue()))
+            if (!documentUrlNode.isMissingNode && !documentUrlNode.isNull) {
+                try {
+                    if (documentUrlNode.isTextual) {
+                        documentenUris.add(documentUrlNode.textValue())
+                    } else if (documentUrlNode.isArray) {
+                        documentUrlNode.forEach { documentenUris.add(it.textValue()) }
+                    } else {
+                        throw RuntimeException("Invalid URL in '/documenten'. ${documentUrlNode.toPrettyString()}")
```
marijnritense, line 129: Same here, I don't think this is specific enough.

zgw/portaaltaak/src/main/kotlin/com/ritense/portaaltaak/PortaalTaakEventListener.kt
```diff
@@ -0,0 +1,178 @@
...
+                    }
+                } catch (e: MalformedURLException) {
+                    throw RuntimeException("Malformed URL in: '/documenten'", e)
```
marijnritense, line 132: Isn't this making it less specific?

zgw/portaaltaak/src/main/kotlin/com/ritense/portaaltaak/PortaalTaakEventListener.kt
```diff
@@ -0,0 +1,178 @@
...
+        )
+        try {
+            processService.startProcess(
+                processDefinitionKey,
+                businessKey,
+                variables
+            )
+        } catch (ex: RuntimeException) {
+            throw NotificatiesException(
+                "Could not start process with definition: $processDefinitionKey and businessKey: $businessKey.\n " +
```
marijnritense, line 160: Should be a more specific exception

zgw/portaaltaak/src/main/kotlin/com/ritense/portaaltaak/PortaaltaakPlugin.kt
```diff
@@ -16,21 +16,254 @@
...
+    )
+    fun completePortaalTaak(delegateExecution: DelegateExecution) {
+        val verwerkerTaakId = delegateExecution.getVariable("verwerkerTaakId") as String
+        val objectenApiPluginId = delegateExecution.getVariable("objectenApiPluginConfigurationId") as String
+        val portaalTaakObjectResourceUrl =
+            URI(delegateExecution.getVariable("portaalTaakObjectResourceUrl") as String)
+        taskService.complete(verwerkerTaakId)
+        val objectenApiPlugin =
+            pluginService.createInstance(PluginConfigurationId(UUID.fromString(objectenApiPluginId))) as ObjectenApiPlugin
+        val portaalTaakMetaObject = objectenApiPlugin.getObject(portaalTaakObjectResourceUrl)
```
marijnritense, line 129: portaalTaakMetaObject -> portaalTaakMetaDataObject

zgw/portaaltaak/src/main/kotlin/com/ritense/portaaltaak/PortaaltaakPlugin.kt
```diff
@@ -16,21 +16,254 @@
...
+    }
+
+    @PluginAction(
+        key = "complete-portaaltaak",
+        title = "Complete Portaaltaak",
+        description = "Complete portal task and update status on Objects Api",
+        activityTypes = [ActivityType.SERVICE_TASK_START]
+    )
+    fun completePortaalTaak(delegateExecution: DelegateExecution) {
+        val verwerkerTaakId = delegateExecution.getVariable("verwerkerTaakId") as String
```
marijnritense, line 122: All of these can be empty, but I think they should always be provided. So if these values are null, we should throw an exception.

zgw/portaaltaak/src/test/kotlin/com/ritense/portaaltaak/PortaaltaakPluginTest.kt
```diff
@@ -0,0 +1,326 @@
...
+        verify(pluginService, times(1)).createInstance(any<Class<ZakenApiPlugin>>(), any())
+        verify(objectManagementService, times(1)).getById(any())
+        verify(delegateTask, times(2)).processInstanceId
+        verify(delegateTask, times(1)).id
+        verify(delegateTask, times(1)).name
+        verify(zaakInstanceLinkService, times(1)).getByDocumentId(any())
+        verify(jsonSchemaDocumentId, times(1)).id
+        verify(processDocumentService, times(2)).getDocumentId(any(), any<DelegateTask>())
+        verify(zakenApiPlugin, times(1)).getZaakRollen(any(), any())
+        verify(valueResolverService, times(1)).resolveValues(any(), any())
```
marijnritense, line 132: What are we testing here? For unit tests, you don't want to test if the code that is written matches exactly what you think it should be. Instead, try to write tests in such a way you can verify the output is what you expect (given certain scenarios).

zgw/portaaltaak/src/main/kotlin/com/ritense/portaaltaak/PortaalTaakEventListener.kt
luis-soares-ritense: here i'm just capturing the exception that startProcess might throw.
in the line bellow i'm appending the exception message which should provide more context on what actually went wrong

zgw/portaaltaak/src/main/kotlin/com/ritense/portaaltaak/PortaalTaakEventListener.kt
luis-soares-ritense: sorry didn't understand the issue here. can you explain?

zgw/portaaltaak/src/main/kotlin/com/ritense/portaaltaak/PortaalTaakEventListener.kt
luis-soares-ritense: properties.get() returns a JsonNode. objectManagement.id is a UUID
not sure how equals works in kotlin

Klaas-Ritense (approved):

PR #546, opened 2023-02-09
Reverted changes to TaakObjectDto

theo-ritense:
Reverted changes to TaakObjectDto because the TaakObjectConnector should not have been changed.

The new json model will be supported using the portaaltaak module

### DIFF
--- a/objects-api/src/main/kotlin/com/ritense/objectsapi/taak/TaakObjectConnector.kt
+++ b/objects-api/src/main/kotlin/com/ritense/objectsapi/taak/TaakObjectConnector.kt
@@ -128,22 +128,9 @@ class TaakObjectConnector(
         formulierId: String?,
         formulierUrl: String?,
     ): TaakObjectDto {
-        val bsn = bsnProvider?.getBurgerServiceNummer(task)
-        val kvk = kvkProvider?.getKvkNummer(task)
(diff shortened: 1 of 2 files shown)


objects-api/src/main/kotlin/com/ritense/objectsapi/taak/TaakObjectConnector.kt
```diff
@@ -128,22 +128,9 @@ class TaakObjectConnector(
...
-        }
-        if (kvk == null && bsn == null) {
-            throw IllegalStateException("Unable to find the 'Initiator' of task '${task.id}'")
-        }
         val taakObject = TaakObjectDto(
-            bsn = bsn,
-            kvk = kvk,
-            identificatie = identificaties,
+            bsn = bsnProvider?.getBurgerServiceNummer(task),
+            kvk = kvkProvider?.getKvkNummer(task),
```
Klaas-Ritense, line 133: Im not understanding the plan here. This format is no longer supported by the portal. Are we going to delete the TaakConnector?

theo-ritense: Quoting Thomas: "abandon it"

Klaas-Ritense (approved):

PR #559, opened 2023-02-16
added context to TaskExtended, which is filled based on the "context" property in process variables.

marijnritense:


core/src/main/java/com/ritense/valtimo/service/CamundaTaskService.java
```diff
@@ -181,6 +182,8 @@ public Page<TaskExtended> findTasksFiltered(
         var parameters = buildTaskFilterParameters(taskFilter);
         Page<TaskExtended> tasks = camundaTaskRepository.findTasks(pageable, parameters);
         if (!tasks.isEmpty()) {
+            tasks.forEach(task -> task.setContext(taskService.getVariable(task.getId(), "context")));
```
theo-ritense, line 185: I feel like "context" should be some constant?

Klaas-Ritense (approved):

PR #565, opened 2023-02-17
Change: portaaltaak other receiver

Klaas-Ritense:


ivo-ritense:
Shouldn't OtherTaakReceiver be cleaned up as well?

ivo-ritense (approved):

PR #575, opened 2023-02-21
Bugfix: fix error specified key was too long

Klaas-Ritense:
```
Specified key was too long; max key length is 3072 bytes [Failed SQL: (1071) ALTER TABLE `some-db`.process_form_association_v2 MODIFY process_definition_key VARCHAR(255)]
```

Solution: The column `process_form_association_v2.id` wasn't used. But after this PR, it is used


theo-ritense:
I think it's okay, but I first want this to be tested in development

marijnritense (approved):

PR #576, opened 2023-02-21
Fix mysql error: Specified key was too long

Klaas-Ritense:


app/gzac/build.gradle
```diff
@@ -76,6 +76,7 @@ dependencies {
 
     // Database
     implementation group: 'org.postgresql', name: 'postgresql', version: postgresqlDriverVersion
+    implementation "mysql:mysql-connector-java"
```
theo-ritense, line 79: Can we do this, license-wise?

Klaas-Ritense: what do you mean?

theo-ritense: It has been added in the past, then removed because of license issues, and now it's added again. I'm lost.

Klaas-Ritense: I've removed it for now

theo-ritense (approved):

PR #585, opened 2023-02-22
No longer remove contexts on each application boot.

nielserkamp-ritense:
**Currently**: Whenever we boot up and run the application the contexts are 'removed' and reset to empty. This means any supporting process with a user task is no longer visible on the dashboard until you 'add' the process to the default context again. This actually also happens on custom made ones too.

**Solution**: Only remove a context process link, when the process is set to 'visibleInMenu'. This should leave others alone and not reset all processes context links anymore.

**Remark**: Though if it is the complete correct solution. As currently you would overwrite any 'manual' adjustments to the context. Might be better to only add, if none exists.


process-document/src/main/java/com/ritense/processdocument/service/impl/CamundaProcessJsonSchemaDocumentDeploymentService.java
```diff
@@ -114,8 +114,8 @@ private void createProcessDocumentLink(String documentDefinitionName, ProcessDoc
         }
 
         contextService.findAll(Pageable.unpaged()).forEach(context -> {
-            context.getProcesses().removeIf(contextProcess -> contextProcess.getProcessDefinitionKey().equals(item.getProcessDefinitionKey()));
             if (Boolean.TRUE.equals(item.getProcessIsVisibleInMenu())) {
+                context.getProcesses().removeIf(contextProcess -> contextProcess.getProcessDefinitionKey().equals(item.getProcessDefinitionKey()));
```
Klaas-Ritense, line 118: This is not going to work when someone tries to disable the `processIsVisibleInMenu` through the auto-deployment file. 
Because when changed to `"processIsVisibleInMenu": false`, the context isn't saved here.

nielserkamp-ritense: But it is not working now either. As I do not get my tasks to show up on the dashboard, even if I have "processIsVisibleInMenu": false because it always removes it. The context menu also has 3 options and not two, so the code is incorrect either way. 

What should the processIsVisibleInMenu actually do?

Klaas-Ritense: Just looked into this. The context menu shows 3 options, but in reality only has 2 options. Which is visible when opening the network tab when saving the contexts.

The `processIsVisibleInMenu = true` is the same as the option `process and tasks`.

I've made a fix that won't in this PR: https://github.com/valtimo-platform/valtimo-backend-libraries/pull/610/files

ivo-ritense: @Klaas-Ritense @nielserkamp-ritense The PR Klaas mentioned has been merged. Does that solve the issue and make this PR obsolete?

PR #600, opened 2023-02-27
Feature/47714 verzoek plugin   mapping verzoek object to gzac document

Klaas-Ritense:


document/src/main/kotlin/com/ritense/document/domain/EveritSchemaAllowsProperty.kt
```diff
@@ -0,0 +1,198 @@
...
+    val remaining = headAndTail[1]
+    val field2 = headAndTail[2]!!
+    return field2.isNotEmpty() && (allowsSchemaProperty(nextToken, remaining)
+            || allowsPatternProperty(nextToken, remaining)
+            || allowsSchemaDependencyProperty(field2)
+            || permitsAdditionalProperties())
+}
+
+private fun ObjectSchema.allowsSchemaProperty(current: String, remaining: String?): Boolean {
+    val currentUnescaped = jSONPointerUnescape(current)
```
theo-ritense, line 51:
```suggestion
    val currentUnescaped = jsonPointerUnescape(current)
```

document/src/main/kotlin/com/ritense/document/domain/EveritSchemaAllowsProperty.kt
```diff
@@ -0,0 +1,198 @@
...
+    val hasSuffix = remaining != null
```
theo-ritense, line 52:
```suggestion
```

document/src/main/kotlin/com/ritense/document/domain/EveritSchemaAllowsProperty.kt
```diff
@@ -0,0 +1,198 @@
...
+    return if (propertySchemas.containsKey(currentUnescaped)) {
+        if (hasSuffix) {
```
theo-ritense, line 54:
```suggestion
        if (remaining != null) {
```

document/src/main/kotlin/com/ritense/document/domain/EveritSchemaAllowsProperty.kt
```diff
@@ -0,0 +1,198 @@
...
+            propertySchemas[currentUnescaped]!!.allowsProperty(remaining!!)
```
theo-ritense, line 55:
```suggestion
            propertySchemas[currentUnescaped]!!.allowsProperty(remaining)
```

document/src/main/java/com/ritense/document/service/DocumentDefinitionService.java
```diff
@@ -67,4 +67,5 @@ public interface DocumentDefinitionService {
 
     boolean isValidJsonPath(JsonSchemaDocumentDefinition definition, String jsonPathExpression);
 
+    void validateJsonPointer(String documentDefinitionName, String jsonPointer);
```
marijnritense, line 70: Just a throught, not something to act on per say:

If we know an exception can occur (by design, in this case because the JsonPointer does not point at an existing property in the document definition) shouldn't we also make throwing the exception a part of the interface? So then it would become:

`void validateJsonPointer(String documentDefinitionName, String jsonPointer) throws ValidationException;`

document/src/main/kotlin/com/ritense/document/domain/EveritSchemaAllowsProperty.kt
```diff
@@ -0,0 +1,198 @@
...
+import org.everit.json.schema.CombinedSchema
+import org.everit.json.schema.JSONPointer
+import org.everit.json.schema.ObjectSchema
+import org.everit.json.schema.ReferenceSchema
+import org.everit.json.schema.Schema
+import org.everit.json.schema.ValidationException
+import org.everit.json.schema.regexp.Regexp
+
+/** This method is a copy from org.everit.json.schema.Schema.definesProperty(.) but returns true when the schema allows additionalProperties */
+fun Schema.allowsProperty(field: String): Boolean {
```
marijnritense, line 29: Can you add a test to validate this method works as intended after the changes made here?

document/src/main/java/com/ritense/document/service/DocumentDefinitionService.java
Klaas-Ritense: I've added it

marijnritense (approved):

PR #605, opened 2023-03-01
#53752 new endpoint to download document from documenten api plugin

Klaas-Ritense:


zgw/documenten-api/src/main/kotlin/com/ritense/documentenapi/DocumentenApiPlugin.kt
```diff
@@ -121,6 +124,46 @@ class DocumentenApiPlugin(
...
+    @PluginAction(
+        key = "download-document",
+        title = "Download document",
+        description = "Download a document from the Documenten API and store it as a temporary document",
+        activityTypes = [ActivityType.SERVICE_TASK_START]
+    )
+    fun downloadInformatieObject(
+        execution: DelegateExecution,
+    ): String {
+        val documentUrl = URI(execution.getVariable(DOCUMENT_URL_PROCESS_VAR) as String)
```
marijnritense, line 136: We know part of this as well from the `url` property that is a part of this plugin. Does it make sense to validate the value of the process variable is a match with this?

Klaas-Ritense: I've added some validation

marijnritense (approved):

PR #612, opened 2023-03-06
Task forms cannot be prefilled in case of multiple processes with the same variable name

marijnritense:
Made prefillTaskVariables actually fill using process variables as well.
Removed prefilling using all processes when a task ID is present.

### DIFF
--- a/form-flow-valtimo/src/main/kotlin/com/ritense/valtimo/formflow/handler/FormFlowStepTypeFormHandler.kt
+++ b/form-flow-valtimo/src/main/kotlin/com/ritense/valtimo/formflow/handler/FormFlowStepTypeFormHandler.kt
@@ -70,7 +70,10 @@ class FormFlowStepTypeFormHandler(
         val document = documentService.get(documentId)
         val documentContent = document.content().asJson() as ObjectNode
 
-        camundaFormAssociationService.prefillProcessVariables(formDefinition, document)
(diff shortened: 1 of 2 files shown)


form-flow-valtimo/src/main/kotlin/com/ritense/valtimo/formflow/handler/FormFlowStepTypeFormHandler.kt
```diff
@@ -70,7 +70,10 @@ class FormFlowStepTypeFormHandler(
         val document = documentService.get(documentId)
         val documentContent = document.content().asJson() as ObjectNode
 
-        camundaFormAssociationService.prefillProcessVariables(formDefinition, document)
+        if (taskInstanceId == null) {
```
Klaas-Ritense, line 73: I just did some testing: A user task form IO isn't being prefilled with 'regular' process variables

marijnritense: What are "regular" process variables?

Klaas-Ritense: never mind, it does seem to work

Klaas-Ritense (approved):

PR #613, opened 2023-03-06
Rename verzoek mapping

ivo-ritense:


zgw/verzoek/src/main/kotlin/com/ritense/verzoek/VerzoekPlugin.kt
```diff
@@ -57,10 +57,10 @@ class VerzoekPlugin(
             .filter { it.copyStrategy == CopyStrategy.SPECIFIED }
             .forEach { property ->
                 property.mapping?.forEach {
-                    if (!it.key.startsWith("doc:")) {
-                        throw IllegalArgumentException("Failed to set mapping. Unknown prefix '${it.key.substringBefore(":")}:'.")
+                    if (!it.target.startsWith("doc:")) {
+                        throw IllegalArgumentException("Failed to set mapping. Unknown prefix '${it.target.substringBefore(":")}:'.")
```
Klaas-Ritense, line 61: Now that you're here anyway :), could you change this to an exception that is visible in the FE?

Because right now, it shows a 500 error without any message in FE

Klaas-Ritense (approved):

PR #626, opened 2023-03-10
#36300 zaken api plugin action - 'set zaakstatus' and 'Create Zaakresultaat'

Klaas-Ritense:
…' & 'Create Zaakresultaat'

### DIFF
--- a/zgw/zaken-api/src/main/kotlin/com/ritense/zakenapi/ZakenApiPlugin.kt
+++ b/zgw/zaken-api/src/main/kotlin/com/ritense/zakenapi/ZakenApiPlugin.kt
@@ -26,6 +26,8 @@ import com.ritense.resource.service.TemporaryResourceStorageService
 import com.ritense.zakenapi.client.LinkDocumentRequest
 import com.ritense.zakenapi.client.ZakenApiClient
 import com.ritense.zakenapi.domain.CreateZaakRequest
+import com.ritense.zakenapi.domain.CreateZaakResultaatRequest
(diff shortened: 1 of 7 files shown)


zgw/zaken-api/src/main/kotlin/com/ritense/zakenapi/ZakenApiPlugin.kt
```diff
@@ -220,6 +223,57 @@ class ZakenApiPlugin(
...
+        description = "Sets the status of a zaak",
+        activityTypes = [ActivityType.SERVICE_TASK_START]
+    )
+    fun setZaakStatus(
+        execution: DelegateExecution,
+        @PluginActionProperty statustypeUrl: URI,
+        @PluginActionProperty statustoelichting: String?,
+    ) {
+        val documentId = UUID.fromString(execution.businessKey)
+        val zaakUrl = zaakUrlProvider.getZaakUrl(documentId)
```
marijnritense, line 238: It could fail on this, right? If there's no zaak that can be found. Should we throw a more specific exception here? Same goes for `createZaakResultaat` as it is using the same method.

Klaas-Ritense: The function `getZaakUrl(..)` will throw an `ZaakInstanceLinkNotFoundException` when there is no zaak found for that document

marijnritense (approved):

PR #633, opened 2023-03-13
#50264 Catalogi API - new endpoint for retrieving Besluittypes

Klaas-Ritense:


zgw/catalogi-api/src/main/kotlin/com/ritense/catalogiapi/CatalogiApiPlugin.kt
```diff
@@ -140,6 +142,27 @@ class CatalogiApiPlugin(
...
     }
 
+    fun getBesluitTypes(zaakTypeUrl: URI): List<Besluittype> {
+        var currentPage = 1
+        var currentResults: Page<Besluittype>?
+        val results = mutableListOf<Besluittype>()
+
+        do {
+            logger.debug { "Getting page of besluit types, page $currentPage for zaaktype $zaakTypeUrl" }
+            currentResults = client.getBesluittypen(
```
marijnritense, line 152: It's `getBesluittypen` here and the method we're in is `getBesluitTypes`. Which of the two is correct?

marijnritense (approved):

PR #659, opened 2023-03-31
Bugfix: /candidate-users doesn't reutnr all users

Klaas-Ritense:
Root cause: the query-param `max` is by default 100

### DIFF
--- a/keycloak-iam/src/main/java/com/valtimo/keycloak/service/KeycloakUserManagementService.java
+++ b/keycloak-iam/src/main/java/com/valtimo/keycloak/service/KeycloakUserManagementService.java
@@ -23,6 +23,14 @@ import com.ritense.valtimo.contract.authentication.model.SearchByUserGroupsCrite
 import com.ritense.valtimo.contract.authentication.model.ValtimoUser;
 import com.ritense.valtimo.contract.authentication.model.ValtimoUserBuilder;
 import com.ritense.valtimo.contract.utils.SecurityUtils;
+import org.keycloak.representations.idm.RoleRepresentation;
+import org.keycloak.representations.idm.UserRepresentation;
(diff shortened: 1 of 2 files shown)


keycloak-iam/src/main/java/com/valtimo/keycloak/service/KeycloakUserManagementService.java
```diff
@@ -23,24 +23,27 @@
...
-import org.slf4j.Logger;
-import org.slf4j.LoggerFactory;
-import org.springframework.data.domain.Page;
-import org.springframework.data.domain.Pageable;
+
 import static com.ritense.valtimo.contract.Constants.SYSTEM_ACCOUNT;
 
 public class KeycloakUserManagementService implements UserManagementService {
     private static final Logger logger = LoggerFactory.getLogger(KeycloakUserManagementService.class);
+    protected static final int MAX_USERS = 1000;
```
theo-ritense, line 46: Isn't this just stretching the problem? @Klaas-Ritense already noticed the current situation was very slow. This will cause it to reach loadbalancer-time-limits.

theo-ritense (dismissed):
Accepted because of the importance, but we need a long term fix.

keycloak-iam/src/main/java/com/valtimo/keycloak/service/KeycloakUserManagementService.java
Klaas-Ritense: I've changed it. 
For the user list on the case page, it now no longer retrieves the roles for every user. 
For the user list in a user-task, it still retrieves the roles for every user, but it now does it for all users in parallel, by using `parallelStream()`

theo-ritense (approved):

PR #662, opened 2023-04-04
Feature/50056 api endpoint for retrieving process link candidates

luis-soares-ritense:


process-link/src/main/kotlin/com/ritense/processlink/domain/SupportedProcessLinkType.kt
```diff
@@ -0,0 +1,23 @@
...
+ *  distributed under the License is distributed on an "AS IS" basis,
+ *  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ *  See the License for the specific language governing permissions and
+ *  limitations under the License.
+ *
+ */
+
+package com.ritense.processlink.domain
+
+interface SupportedProcessLinkType {
```
theo-ritense, line 20: Please postfix this interface name and it's implementations with Handler. Otherwise, it's as if they are types of itself.

PR #668, opened 2023-04-12
#61006 support prefilled forms when using process links

Klaas-Ritense:
Changes:
- Moved `FormProcessLink` stuff out of deprecated module `form-link`
- Moved all functions in `SubmissionTransformerService` out of deprecated module `form-link` to new `PrefillFormService`
- Moved all prefill functions in `CamundaFormAssociationService` out of deprecated module `form-link` to new `PrefillFormService`
- Create `FormProcessLinkTaskProvider`


form/src/main/kotlin/com/ritense/form/domain/FormTaskOpenResultProperties.kt
```diff
@@ -0,0 +1,25 @@
...
+ */
+
+package com.ritense.form.domain
+
+import com.fasterxml.jackson.databind.JsonNode
+import java.util.UUID
+
+data class FormTaskOpenResultProperties(
+    val formDefinitionId: UUID,
+    val prefilledForm: JsonNode,
```
theo-ritense, line 24: We might also want to use this class for returning a form for start-events, where prefilling will not be done. Also, a prefilled form is still just a form-definition. 

rename to `formDefinition` ?

form-link/src/main/kotlin/com/ritense/formlink/repository/FormProcessLinkRepository.kt
```diff
@@ -16,7 +16,7 @@
 
 package com.ritense.formlink.repository
 
-import com.ritense.formlink.domain.FormProcessLink
+import com.ritense.form.domain.FormProcessLink
```
theo-ritense, line 19: Is this still used? I can't find any usage. And if it's used, shouldn't this be part of the `form` module?

Klaas-Ritense: It was unused so I removed it

theo-ritense (approved):

PR #676, opened 2023-04-17
Check for prefill property on prefilling form fields.

theo-ritense:


form-link/src/main/kotlin/com/ritense/formlink/domain/impl/submission/formfield/UploadField.kt
Klaas-Ritense, line 1: Lets first test it on dev

theo-ritense: Good idea.

PR #678, opened 2023-04-18
59998: Processlink autodeployment

theo-ritense:


Klaas-Ritense (approved):
lgtm.
Would be nice if we convert all te plugin-actions in the `/app/gzac/ApplicationReadyEventListener` to a json config file

theo-ritense:
@Klaas-Ritense We can't (yet), as the plugin-processlinks need a reference to a plugin configuration.

PR #684, opened 2023-04-21
Fix: failed to create process link due to missing property 'processLinkType'

Klaas-Ritense:
Solution: The property 'processLinkType' is now optional. 

For example: An autodeployment like this works now:
```
[{
        "activityId": "my-service-task",
        "activityType": "bpmn:ServiceTask:start",
        "someValue": "test"
}]
```

But this also works and gives the same result:

```
[{
        "activityId": "my-service-task",
        "activityType": "bpmn:ServiceTask:start",
        "processLinkType": "test",
        "someValue": "test"
}]
```

### DIFF
--- a/process-link/src/test/kotlin/com/ritense/processlink/domain/CustomProcessLinkUpdateRequestDtoTest.kt
+++ b/process-link/src/test/kotlin/com/ritense/processlink/domain/CustomProcessLinkUpdateRequestDtoTest.kt
@@ -20,13 +20,13 @@ import com.fasterxml.jackson.module.kotlin.jacksonObjectMapper
 import com.fasterxml.jackson.module.kotlin.readValue
 import com.ritense.processlink.domain.CustomProcessLink.Companion.PROCESS_LINK_TYPE_TEST
 import com.ritense.processlink.web.rest.dto.ProcessLinkUpdateRequestDto
-import java.util.UUID
 import org.hamcrest.MatcherAssert.assertThat
 import org.hamcrest.Matchers.equalTo
 import org.hamcrest.Matchers.instanceOf
 import org.junit.jupiter.api.Test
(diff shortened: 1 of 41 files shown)


theo-ritense:
This re-introduces the problem where one could do the following:
```
[{
        "activityId": "my-service-task",
        "activityType": "bpmn:ServiceTask:start",
        "processLinkType": "form-flow",
        "someValue": "test"
}]
```

Which would cause the link to be processed by the form-flow mapper instead of test. There's nothing that validates the input this way.

theo-ritense:
We might be able to fix this using as custom TypeIdResolver which defaults to 'plugin':

https://stackoverflow.com/questions/34814664/jackson-custom-typeid-resolver

process-link/src/test/kotlin/com/ritense/processlink/domain/CustomProcessLinkUpdateRequestDtoTest.kt
Klaas-Ritense, line 1: @theo-ritense your example is not correct. Field existence is validated first. After that, it will fallback on the `processLinkType` property.

For example:
[{
        "activityId": "my-service-task",
        "activityType": "bpmn:ServiceTask:start",
        "processLinkType": "form-flow",
        "someValue": "test"
}]
This JSON will be processed by the 'test' mapper. Because of the existence of field `someValue`. But if that mapper doesn't exist, the JSON will be processed by the formFlowMapper

theo-ritense: Sorry, my comment wasn't very clear. I'm not talking about the `ObjectMapper`. I meant the `ProcessLinkMapper`. That mapper is selected based on the value of `processLinkType`

PR #691, opened 2023-04-26
#61681 fix: Searching Objects

Klaas-Ritense:


zgw/object-management/src/main/kotlin/com/ritense/objectmanagement/service/ObjectManagementService.kt
```diff
@@ -121,35 +129,15 @@ class ObjectManagementService(
...
-                            searchField.fieldType,
-                            listOf(otherFilter.rangeFrom?.value, otherFilter.rangeTo?.value),
-                            searchField.dataType
-                        )
-                    }
-                }
+                .flatMap { otherFilter -> mapToObjectSearchParameters(searchField, otherFilter) }
         }
 
         val searchString = ObjectSearchParameter.toQueryParameter(searchDtoList)
```
theo-ritense, line 143: I just noticed we don't do sanitation when converting values to query parameters. I think we should url-encode them.

Klaas-Ritense: I just looked it up: a minimal amount of url encoding is done by default in the webclient when adding this searchstring as a query parameter. [docs](https://docs.spring.io/spring-framework/docs/current/javadoc-api/org/springframework/web/util/UriBuilder.html#queryParam(java.lang.String,java.lang.Object...))

marijnritense (dismissed):

theo-ritense (approved):

search/src/main/kotlin/com/ritense/search/domain/DataType.kt
```diff
@@ -20,7 +20,7 @@ enum class DataType {
     TEXT,
     NUMBER,
     DATE,
-    DATE_TIME,
+    DATETIME,
```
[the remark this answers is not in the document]
Klaas-Ritense: To make it the same to the DataType.DATETIME that is used by [case searches ](https://github.com/valtimo-platform/valtimo-backend-libraries/blob/main/document/src/main/java/com/ritense/document/domain/impl/searchfield/SearchFieldDataType.java)

marijnritense (approved):

PR #706, opened 2023-05-08
Added required variables to endpoint to start form flow

ivo-ritense:


theo-ritense (commented):
I'm confused. You say it's required, but the code tells me it's optional.

ivo-ritense:
> I'm confused. You say it's required, but the code tells me it's optional.

That's because the arguments are required in different situation. e.g. documentId is used to prefill the data.

theo-ritense (approved):

PR #718, opened 2023-05-11
Feature/pbac

marijnritense:
Includes support for roles, interface for AuthorizationService, and using users roles, retrieves corresponding permissions from the DB. Contains a lot of TODOs that have to be picked up in other stories.


build.gradle
```diff
@@ -89,7 +89,7 @@ allprojects {
                 mavenBom "org.junit:junit-bom:${junitVersion}"
             }
             dependencies {
-                dependency "org.yaml:snakeyaml:${snakeYamlVersionOverride}"
+//                dependency "org.yaml:snakeyaml:${snakeYamlVersionOverride}"
```
Klaas-Ritense, line 92: The snakeyaml dependency version from the gradle.properties, should be used here

build.gradle
```diff
@@ -89,7 +89,6 @@ allprojects {
                 mavenBom "org.junit:junit-bom:${junitVersion}"
             }
             dependencies {
-                dependency "org.yaml:snakeyaml:${snakeYamlVersionOverride}"
                 dependency "io.netty:netty-codec:${nettyCodecVersionOverride}"
```
theo-ritense, line 92:
```suggestion
                dependency "org.yaml:snakeyaml:${snakeYamlVersionOverride}"
                dependency "io.netty:netty-codec:${nettyCodecVersionOverride}"
```

build.gradle
theo-ritense: That override broke some things because it was incompatible with, I believe: liquibase. But I'll check.

plugin/src/main/kotlin/com/ritense/plugin/configuration/PluginAutoConfiguration.kt
```diff
@@ -116,7 +116,7 @@ class PluginAutoConfiguration {
         pluginProcessLinkRepository: PluginProcessLinkRepository,
         @Lazy pluginFactories: List<PluginFactory<*>>,
         objectMapper: ObjectMapper,
-        @Lazy valueResolverService: ValueResolverService,
+        valueResolverService: ValueResolverService,
```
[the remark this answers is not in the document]
theo-ritense: Too be fair: it shouldn't. It was breaking things because we were using the wrong version of the valueresolver module.

build.gradle
theo-ritense: @Klaas-Ritense Added it again.

notes/src/main/kotlin/com/ritense/note/web/rest/NoteResource.kt
```diff
@@ -60,9 +60,9 @@ class NoteResource(
 
         val jsonSchemaDocumentId = JsonSchemaDocumentId.existingId(documentId)
 
-        if (!documentService.currentUserCanAccessDocument(jsonSchemaDocumentId)) {
+/*        if (!documentService.currentUserCanAccessDocument(jsonSchemaDocumentId)) {
             return ResponseEntity.status(HttpStatus.FORBIDDEN).build()
-        }
+        }*/
```
Klaas-Ritense, line 65: This can be remove now, right?

Klaas-Ritense (approved):

PR #722, opened 2023-05-15
#62954 move plugin-process-link to plugin-valtimo package

Klaas-Ritense:


theo-ritense (commented):
Did we think this one through? I think the goal is to make the plugin module independent of Valtimo again, but by doing this we make the process-links dependant on plugins...

In hindsight, shouldn't this be a separate (glue) module?

PR #725, opened 2023-05-16
Added support for doc: and pv: fields for form flow expressions when …

ivo-ritense:
…starting a case

### DIFF
--- a/form-flow-valtimo/src/main/kotlin/com/ritense/valtimo/formflow/common/ValtimoFormFlow.kt
+++ b/form-flow-valtimo/src/main/kotlin/com/ritense/valtimo/formflow/common/ValtimoFormFlow.kt
@@ -19,17 +19,30 @@ package com.ritense.valtimo.formflow.common
 import com.fasterxml.jackson.core.JsonPointer
 import com.fasterxml.jackson.databind.JsonNode
 import com.fasterxml.jackson.databind.ObjectMapper
+import com.fasterxml.jackson.module.kotlin.jacksonObjectMapper
+import com.fasterxml.jackson.module.kotlin.readValue
+import com.ritense.document.domain.impl.JsonSchemaDocumentId
+import com.ritense.document.domain.impl.request.NewDocumentRequest
+import com.ritense.document.service.DocumentService
+import com.ritense.formflow.domain.instance.FormFlowInstanceId
 import com.ritense.formflow.expression.FormFlowBean
-import com.ritense.valtimo.contract.json.Mapper
+import com.ritense.formflow.service.FormFlowService
+import com.ritense.processdocument.domain.impl.request.StartProcessForDocumentRequest
+import com.ritense.processdocument.service.ProcessDocumentService
 import com.ritense.valueresolver.ValueResolverService
 import org.camunda.bpm.engine.TaskService
 import org.springframework.transaction.annotation.Transactional
+import java.util.Objects
+import java.util.UUID
 
 @FormFlowBean
 open class ValtimoFormFlow(
     private val taskService: TaskService,
     private val objectMapper: ObjectMapper,
     private val valueResolverService: ValueResolverService,
+    private val formFlowService: FormFlowService,
+    private val processDocumentService: ProcessDocumentService,
+    private val documentService: DocumentService
 ) {
 
     /**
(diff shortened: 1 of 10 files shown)


value-resolver/src/main/kotlin/com/ritense/valueresolver/ValueResolverFactory.kt
```diff
@@ -90,4 +91,24 @@ interface ValueResolverFactory {
...
      */
     fun handleValues(processInstanceId: String, variableScope: VariableScope?, values: Map<String, Any>)
+
+    /**
+     * Handle values for a case where a process is not relevant or present in the current context.
+     *
+     * @param documentId The id of the document these values belong to
+     * @param values The values to handle. i.e. mapOf(doc:add:/firstname to John)
+     */
+    fun handleValues(documentId: UUID, values: Map<String, Any>) {
```
theo-ritense, line 101: Maybe add a @Deprecation notice, so we can refactor it in the next major?

value-resolver/src/main/kotlin/com/ritense/valueresolver/ValueResolverFactory.kt
```diff
@@ -90,4 +91,24 @@ interface ValueResolverFactory {
...
+        //empty default method for backwards compatibility
+    }
+
+    /**
+     * Processes and tranforms values for use externally. This is used when case or process don't exist yet. For example
+     * wehen creating a new case.
+     *
+     * @param values The values to handle. i.e. mapOf(doc:add:/firstname to John)
+     */
+    fun preProcessValuesForNewCase(values: Map<String, Any>): Any {
```
theo-ritense, line 111: Maybe add a @Deprecation notice, so we can refactor it in the next major?

value-resolver/src/main/kotlin/com/ritense/valueresolver/ValueResolverFactory.kt
```diff
@@ -90,4 +91,24 @@ interface ValueResolverFactory {
      * @param values The values to handle. i.e. mapOf(doc:add:/firstname to John)
      */
     fun handleValues(processInstanceId: String, variableScope: VariableScope?, values: Map<String, Any>)
```
theo-ritense, line 93: Maybe add a @Deprecation notice, so we can refactor it in the next major?

form-flow-valtimo/src/main/kotlin/com/ritense/valtimo/formflow/common/ValtimoFormFlow.kt
```diff
@@ -75,6 +88,97 @@ open class ValtimoFormFlow(
...
+        ).toString()
+
+        val submission = jacksonObjectMapper().readValue<JsonNode>(formFlowInstance.getSubmissionDataContext())
+        val submissionValues = submissionSavePath.entries.associate { it.key to getValue(submission, it.value) }
+        valueResolverService.handleValues(UUID.fromString(documentId), submissionValues)
+        val submittedByType = valueResolverService.preProcessValuesForNewCase(submissionValues)
+
+        val processDefinitionKey = getRequiredAdditionalProperty(formFlowInstance.getAdditionalProperties(), "processDefinitionKey").toString()
+        val startProcessForDocumentRequest = StartProcessForDocumentRequest(
+            JsonSchemaDocumentId.existingId(UUID.fromString(getRequiredAdditionalProperty(formFlowInstance.getAdditionalProperties(), "documentId").toString())),
```
Klaas-Ritense, line 162:
```suggestion
            JsonSchemaDocumentId.existingId(UUID.fromString(documentId)),
```

value-resolver/src/main/kotlin/com/ritense/valueresolver/ValueResolverFactory.kt
marijnritense: I think we should do that when we introduce the new way. That way, people don't see the notice without us providing an alternative.

form-flow-valtimo/src/main/kotlin/com/ritense/valtimo/formflow/common/ValtimoFormFlow.kt
```diff
@@ -75,6 +88,97 @@ open class ValtimoFormFlow(
...
+                    startProcessForDocumentResult.errors().joinToString(separator = "\n - ")
+            )
+        }
+    }
+
+    /**
+     * Starts a process for an existing case
+     */
+    @Transactional
+    open fun startSupportingProcess(
```
Klaas-Ritense, line 145: Is it possible to combine this function with the `startCase(..)` function above? So that implementation teams don't have to create 2 identical form-flows that only differ in the `onComplete`

theo-ritense (approved):

PR #730, opened 2023-05-24
fix: error when escaped unicode characters are split by buffering

Klaas-Ritense:


document-generation/smartdocuments/src/main/kotlin/com/ritense/smartdocuments/client/SmartDocumentsClient.kt
```diff
@@ -201,9 +201,17 @@ class SmartDocumentsClient(
...
+            try {
+                write(
+                    documentDataOutWriter,
+                    inputStream,
+                    parsedResponse.documentDataStart,
+                    parsedResponse.documentDataEnd
+                )
+            } finally {
+                documentDataOutWriter.close()
+                inputStream.close()
```
theo-ritense, line 213: The method that created the stream is responsible for closing it. inputstream is provided, so should not be closed here.
And it Kotlin, you can use `.use`. See https://kotlinlang.org/api/latest/jvm/stdlib/kotlin.io/use.html

theo-ritense (approved):

PR #732, opened 2023-05-25
#66318 add support for operators on field conditions

Klaas-Ritense:


authorization/src/main/resources/config/liquibase/changelog/20230424-create-authorization-table.xml
```diff
@@ -58,4 +58,8 @@
                                  referencedColumnNames="key"/>
     </changeSet>
 
+    <changeSet author="Ritense" id="5">
+        <delete tableName="permission"/>
```
marijnritense, line 62: Uh... what?

authorization/src/main/resources/config/liquibase/changelog/20230424-create-authorization-table.xml
```diff
@@ -58,4 +58,8 @@
                                  referencedColumnNames="key"/>
     </changeSet>
 
+    <changeSet author="Ritense" id="5">
```
ivo-ritense, line 61: Why do we have to do this?

ivo-ritense: Is it because it wouldn't deserialize without the new operators in there?

Klaas-Ritense: Yes. People that have been developing locally will get an error if this permission table isn't emptied. Because the JSON inside this table is missing the new 'operator' field. Which is now mandatory

ivo-ritense (approved):

authorization/src/main/resources/config/liquibase/changelog/20230424-create-authorization-table.xml
Klaas-Ritense: Not sure what you mean there. Any permission that existed before this changeset were invalid. And permissions that are loaded on startup are always executed after this changeset

PR #738, opened 2023-05-31
#56899 enable implementations to use security smoke integration test

Klaas-Ritense:


theo-ritense:
Just ran the BesluitSecuritySmokeIntegrationTest locally, and no tests where found while there are endpoints in that module. I'll check.

test-utils-common/src/main/kotlin/com/ritense/valtimo/web/rest/SecuritySmokeIntegrationTest.kt
```diff
@@ -0,0 +1,150 @@
...
+@ExtendWith(SpringExtension::class)
+@SpringBootTest
+@AutoConfigureMockMvc
+@Tag("security")
+abstract class SecuritySmokeIntegrationTest(
+    private val basePackageName:String? = null,
+    private val ignoredPathPatterns: Set<String> = setOf()
+) {
+    @MockBean
+    var userManagementService: UserManagementService? = null
```
Klaas-Ritense, line 51: Something like this should be possible:
```suggestion
    lateinit var userManagementService: UserManagementService
```

theo-ritense (approved):

PR #740, opened 2023-05-31
feat: 66812 automatic assignment of tasks to case handler

anna-ritense:
Case Settings now allow for automatic assignments of user task (new and existing) and unassignment of existing assigned tasks to the case assignee via an additional property in CaseSettings.

### DIFF
--- a/case/src/test/kotlin/com/ritense/case/web/dto/CaseSettingsDtoTest.kt
+++ b/case/src/test/kotlin/com/ritense/case/web/dto/CaseSettingsDtoTest.kt
@@ -30,4 +30,23 @@ class CaseSettingsDtoTest {
         assertNull(caseSettingsDto.canHaveAssignee)
         assertFalse(updatedCaseSettings.canHaveAssignee)
     }
+
+    @Test
+    fun `should set autoAssignTasks to false when canHaveAssignee is set to false`() {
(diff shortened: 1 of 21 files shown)


case/src/test/kotlin/com/ritense/case/web/dto/CaseSettingsDtoTest.kt
```diff
@@ -30,4 +30,23 @@ class CaseSettingsDtoTest {
...
+        val caseSettingsDto = CaseSettingsDto(
+            canHaveAssignee = false
+        )
+        val updatedCaseSettings = caseSettingsDto.update(currentCaseSettings)
+
+        assertTrue(currentCaseSettings.canHaveAssignee)
+        assertTrue(currentCaseSettings.autoAssignTasks)
+        assertNull(caseSettingsDto.autoAssignTasks)
+        assertFalse(updatedCaseSettings.canHaveAssignee)
+        assertFalse(updatedCaseSettings.canHaveAssignee)
```
marijnritense, line 50: You're asserting the same thing twice here, this should probably be autoAssignTasks.

```suggestion
        assertFalse(updatedCaseSettings.autoAssignTasks)
```

marijnritense (approved):

PR #747, opened 2023-06-08
Feature/pbac actions

ivo-ritense:


authorization/src/main/kotlin/com/ritense/authorization/ValtimoAuthorizationService.kt
```diff
@@ -59,6 +61,14 @@ class ValtimoAuthorizationService(
         } as AuthorizationEntityMapper<FROM, TO>?)?: throw AccessDeniedException("No entity mapper found for given arguments.")
     }
 
+    override fun <T : Any> getAvailableActionsForResource(clazz: Class<T>): List<Action<T>> {
+        return actionProviders
+            .filter { (it.javaClass.genericInterfaces[0] as ParameterizedType).actualTypeArguments[0].equals(clazz) }
```
marijnritense, line 66: Doesn't quite sit right with me, but it is what it is...

PR #749, opened 2023-06-08
#69531 Fix: plugin deployment problems

Klaas-Ritense:
Solves:
- Error: ```"invalidParams":[{"name":"documentatieLink","code":"null","reason":"Dit veld mag niet leeg zijn."}]```
- Error when running plugin events for a plugin that already exist.
- Error when decrypting plugin secret that hasn't been encrypted yet due to it being in a Transaction.

### DIFF
--- a/plugin/src/main/kotlin/com/ritense/plugin/autodeployment/PluginAutoDeploymentEventListener.kt
+++ b/plugin/src/main/kotlin/com/ritense/plugin/autodeployment/PluginAutoDeploymentEventListener.kt
@@ -52,11 +52,11 @@ class PluginAutoDeploymentEventListener(
                 try {
                     createPluginConfigurations(resource)
                 } catch (e: Exception) {
-                    logger.error(e) { "Error while deploying plugin" }
+                    logger.error(e) { "Error while deploying plugin configuration" }
                 }
             }
         } catch (e: Exception) {
-            logger.error(e) { "Error while deploying plugins" }
+            logger.error(e) { "Error while deploying plugin configurations" }
         }
     }
 
--- a/plugin/src/main/kotlin/com/ritense/plugin/autodeployment/PluginAutoDeploymentEventListener.kt
+++ b/plugin/src/main/kotlin/com/ritense/plugin/autodeployment/PluginAutoDeploymentEventListener.kt
@@ -21,7 +21,6 @@ import com.fasterxml.jackson.databind.ObjectMapper
 import com.fasterxml.jackson.databind.node.ArrayNode
 import com.fasterxml.jackson.databind.node.ObjectNode
 import com.fasterxml.jackson.module.kotlin.treeToValue
-import com.ritense.plugin.domain.PluginConfiguration
 import com.ritense.plugin.service.PluginService
 import mu.KLogger
 import mu.KotlinLogging
@@ -34,7 +33,6 @@ import org.springframework.core.io.ResourceLoader
 import org.springframework.core.io.support.ResourcePatternUtils
 import org.springframework.transaction.annotation.Transactional
 import java.io.IOException
-import java.lang.Exception
 
 @Transactional
 class PluginAutoDeploymentEventListener(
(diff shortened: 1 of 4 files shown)


plugin/src/main/kotlin/com/ritense/plugin/service/PluginService.kt
```diff
@@ -113,15 +113,22 @@ class PluginService(
...
         val pluginDefinition = pluginDefinitionRepository.getById(deploymentDto.pluginDefinitionKey)
+        if (deploymentDto.id != null && pluginConfigurationRepository.existsById(
+                PluginConfigurationId.existingId(deploymentDto.id)
+            )
+        ) {
+            deletePluginConfiguration(PluginConfigurationId.existingId(deploymentDto.id))
+        }
         validateProperties(deploymentDto.properties!!, pluginDefinition)
-        plugin = pluginConfigurationRepository.save(
+        val pluginId = pluginConfigurationRepository.saveAndFlush(
```
marijnritense, line 123: Since all plugin events are ran afterwards, if one of those calls fails it should really just not save the plugin configuration to the database. I think doing a save and flush here just introduces this as a new issue.

Klaas-Ritense: When a plugin-event throws an error, the plugin-configuration is deleted

plugin/src/main/kotlin/com/ritense/plugin/autodeployment/PluginAutoDeploymentEventListener.kt
```diff
@@ -51,7 +49,11 @@ class PluginAutoDeploymentEventListener(
...
             for (resource in resources) {
-                createPluginConfigurations(resource)
+                try {
+                    createPluginConfigurations(resource)
+                } catch (e: Exception) {
+                    logger.error(e) { "Error while deploying plugin" }
+                }
             }
         } catch (e: Exception) {
             logger.error(e) { "Error while deploying plugins" }
```
theo-ritense, line 59: Error while loading plugin configurations

plugin/src/main/kotlin/com/ritense/plugin/service/PluginService.kt
theo-ritense: So, first we delete an existing plugin configuration. Then we run the events. When the events fail, we are left without a pluginconfiguration where we used to have one.

Klaas-Ritense: Not sure what you're suggesting here.

Most of the change are made for OpenNotifications. Because OpenNotifications tries to create an abonnement on the plugin-create-event and delete de abonnement on a plugin-delete-event. Which means that the plugin-events are a mandatory part every time the plugin-configuration is updated. Which wasn't done properly before.

However, the `saveAndFlush` is there for a different reason. The secret of a plugin-configuration is only encrypted when the entity is persisted. Which basically solved [this](https://ritensebv.slack.com/archives/G01Q25SVBHN/p1684755851723469) error that we were seeing on every startup

marijnritense: At line 120 we delete a plugin configuration with that ID already exists in the database. Then, a new plugin configuration is created, and a flush occurs. This means the previous plugin configuration is now removed, and the new one has been saved. If then an even fails, this new plugin configuration is deleted from the database. This means now no plugin configuration with that ID can be found in the database, which is not what you would expect if a deployment fails.

Klaas-Ritense: Flushing should not affect transaction management at all. When transaction is rolled back, all changes made to the database state (i.e. changes flushed by Hibernate) inside that transaction are rolled back.

Klaas-Ritense: I agree that this peace of code isn't particularly beautiful. But this is the only way I could find to have the secret encryption and decryption happen in the correct order. Without making a breaking change, that is.

marijnritense (approved):

PR #756, opened 2023-06-14
#68276 Importing permissions

Klaas-Ritense:


authorization/src/main/kotlin/com/ritense/authorization/deployment/PermissionDeployer.kt
```diff
@@ -0,0 +1,53 @@
...
+            permissionChangeset.permissions
+        )
+    }
+
+    override fun deploy(content: String) {
+        val permissionChangeset = objectMapper.readValue<PermissionChangeset>(content)
+        permissionChangeset.permissions
+            .map { it.toPermission() }
+            .forEach { permission ->
+                permissionRepository.deleteAllByResourceTypeAndActionAndRoleKey(
```
marijnritense, line 45: The combination of `resourceType`, `action` and `roleKey` is not unique. I don't think we should be deleting anything from the database at this step.

core/src/main/kotlin/com/ritense/valtimo/importchangelog/ImportChangelogDeploymentService.kt
```diff
@@ -0,0 +1,121 @@
...
+import org.springframework.core.io.ResourceLoader
+import org.springframework.core.io.support.ResourcePatternUtils
+import org.springframework.transaction.annotation.Transactional
+import java.io.IOException
+import java.text.Normalizer
+import java.time.Instant
+import kotlin.jvm.optionals.getOrNull
+
+@Transactional
+class ImportChangelogDeploymentService(
```
marijnritense, line 38: Thinking about the future: Let's say we wanted to do this for Form Flow as well, how would we do this without adding a dependency on core to the module?

core/src/main/kotlin/com/ritense/valtimo/importchangelog/ImportChangelogDeploymentService.kt
```diff
@@ -0,0 +1,121 @@
...
+    private val changesetDeployers: List<ChangesetDeployer>,
+) {
+
+    @EventListener(ApplicationReadyEvent::class)
+    fun deployAll() {
+        logger.info { "Running import deployer" }
+        loadResources().forEach { resource ->
+            val filename = getFilename(resource)
+            val changesetDeployer = getChangesetDeployer(resource)
+            if (changesetDeployer != null) {
```
marijnritense, line 51: Not sure if a silent fail is the right decision here. Let's discuss tomorrow.

core/src/main/kotlin/com/ritense/valtimo/importchangelog/ImportChangelogDeploymentService.kt
```diff
@@ -0,0 +1,121 @@
...
+@Transactional
+class ImportChangelogDeploymentService(
+    private val resourceLoader: ResourceLoader,
+    private val importChangesetRepository: ImportChangesetRepository,
+    private val objectMapper: ObjectMapper,
+    private val changesetDeployers: List<ChangesetDeployer>,
+) {
+
+    @EventListener(ApplicationReadyEvent::class)
+    fun deployAll() {
```
marijnritense, line 46: When do tables get cleared? I looked around a bit and couldn't find it. As per the story:

> Application property that deletes the contents of the checksum table and the permissions table when set to true. Set to false by default.

core/src/main/kotlin/com/ritense/valtimo/importchangelog/ImportChangelogDeploymentService.kt
```diff
@@ -0,0 +1,121 @@
...
+    private fun getTypeFromFileName(filename: String?): String? {
+        return if (filename == null) {
+            null
+        } else {
+            Regex(".+\\.(.+)\\.json").find(filename)?.groupValues?.get(1)
+        }
+    }
+
+    @Throws(IOException::class)
+    private fun loadResources() = ResourcePatternUtils.getResourcePatternResolver(resourceLoader).getResources(PATH)
```
marijnritense, line 114: Maybe pattern matching to actual resources should be done by the individual deployers.

authorization/src/main/kotlin/com/ritense/authorization/deployment/PermissionDeployer.kt
Klaas-Ritense: A delete statement is necessary to be able to make changes in the permissions

core/src/main/kotlin/com/ritense/valtimo/importchangelog/ImportChangelogDeploymentService.kt
Klaas-Ritense: By using the `contract` module. I've done the same in this PR for the `authorization` module. Because the `authorization` module also doesn't have a dependency on `core`

But you could also add `core` to `form-flow-valtimo`

core/src/main/kotlin/com/ritense/valtimo/importchangelog/ImportChangelogDeploymentService.kt
Klaas-Ritense: It's a must. Because the `loadResources()` loads more that just the permissions and the roles. It also loads `.processlink.json` and `.schema.json` which we dont want to deploy

core/src/main/kotlin/com/ritense/valtimo/importchangelog/ImportChangelogDeploymentService.kt
```diff
@@ -0,0 +1,121 @@
...
+        return changesetDeployers.find { it.getType() == type }
+    }
+
+    private fun saveChangeset(changesetId: String, filename: String, md5sum: String) {
+        importChangesetRepository.save(
+            ImportChangeset(
+                id = changesetId,
+                filename = filename,
+                dateExecuted = Instant.now(),
+                orderExecuted = importChangesetRepository.getNextOrderExecuted() ?: 0,
```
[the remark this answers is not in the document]
Klaas-Ritense: Yes it does. The reason to have a `dateexecuted` and a `orderexecuted` is simply because liquibase has the same columns

core/src/main/kotlin/com/ritense/valtimo/importchangelog/ImportChangelogDeploymentService.kt
Klaas-Ritense: But is we do that, how can we support multiple permission files?
right now we have: `document.permission.json` and `note.permission.json`

marijnritense: This is why I think this should be done by the individual deployers. Have them look at the property, and if it's set then we delete the tables that will be imported (and the relevant records from the changeset table). So in this case, it's about permissions, which means the permissions table gets cleared first. After that, all .permission.json files get imported.

core/src/main/kotlin/com/ritense/valtimo/importchangelog/ImportChangelogDeploymentService.kt
Klaas-Ritense: changed

core/src/main/kotlin/com/ritense/valtimo/importchangelog/ImportChangelogDeploymentService.kt
Klaas-Ritense: changed

authorization/src/main/kotlin/com/ritense/authorization/autoconfigure/AuthorizationAutoConfiguration.kt
```diff
@@ -91,4 +96,28 @@ class AuthorizationAutoConfiguration(
...
+        permissionRepository: PermissionRepository,
+        changelogService: ChangelogService,
+        @Value("\${valtimo.pbac.clear-tables:false}") clearTables: Boolean
+    ): PermissionDeployer {
+        return PermissionDeployer(objectMapper, permissionRepository, changelogService, clearTables)
+    }
+
+    @Bean
+    @ConditionalOnMissingBean(RoleDeployer::class)
+    @Order(2)
```
marijnritense, line 113: Shouldn't the order numbers be really high? So they always run first.

app/gzac/src/main/resources/config/pbac/1-all.role.json
marijnritense, line 1: Any reason why we're including numbers in these filenames?

Klaas-Ritense: Both the number in the filename and the `@Order(2)` on the beans are there to make sure the queries are executed in this specific order:
1. Delete all permissions
2. Delete all roles
3. Create new roles
4. Create new permissions

That order is needed because of the FK constraint between roles and permissions

changelog/src/main/kotlin/com/ritense/valtimo/changelog/service/ChangelogDeployer.kt
```diff
@@ -0,0 +1,65 @@
...
+
+    @EventListener(ApplicationReadyEvent::class)
+    fun deployAll() {
+        logger.info { "Running deployer" }
+
+        changesetDeployers.forEach { it.before() }
+
+        changesetDeployers
+            .flatMap { changesetDeployer ->
+                changelogService.loadResources(changesetDeployer.getPath()).map { Pair(changesetDeployer, it) }
```
marijnritense, line 39: I don't think there's a need to do this in a flatmap. The order is already written down with the `@Order()` annotation. So instead you could to something like:

```
.forEach { (changesetDeployer) ->
    val changelogs = changelogService.loadResources(changesetDeployer.getPath())
   // for each changelog, import the file
}
```

changelog/src/main/kotlin/com/ritense/valtimo/changelog/service/ChangelogDeployer.kt
```diff
@@ -0,0 +1,65 @@
...
+class ChangelogDeployer(
+    private val changelogService: ChangelogService,
+    private val changesetDeployers: List<ChangesetDeployer>,
+) {
+
+    @EventListener(ApplicationReadyEvent::class)
+    fun deployAll() {
+        logger.info { "Running deployer" }
+
+        changesetDeployers.forEach { it.before() }
```
marijnritense, line 35: Shouldn't forEach be done in reverse?

app/gzac/src/main/resources/config/pbac/1-all.role.json
marijnritense: This feels really fragile. I propose a slightly different way of doing it below. The highlights: Deleting and creating should be done by following the `@Order()`. For the deletion, this should happen in the opposite order, and for creating in the normal order. Per deployer, files have to be autodeployed. If we have it be depending on the filename, then it becomes more difficult to predict. After all, roles should always be deployed before permissions. No matter what the filename is.

authorization/src/main/kotlin/com/ritense/authorization/autoconfigure/AuthorizationAutoConfiguration.kt
marijnritense, line 113:
```suggestion
    @Order(1)
```

authorization/src/main/kotlin/com/ritense/authorization/autoconfigure/AuthorizationAutoConfiguration.kt
```diff
@@ -91,4 +96,28 @@ class AuthorizationAutoConfiguration(
     fun <T: Any> denyAuthorizationSpecificationFactory(): AuthorizationSpecificationFactory<T> {
         return DenyAuthorizationSpecificationFactory()
     }
+    @Bean
+    @ConditionalOnMissingBean(PermissionDeployer::class)
+    @Order(1)
```
marijnritense, line 101:
```suggestion
    @Order(2)
```

app/gzac/src/main/resources/config/pbac/1-all.role.json
Klaas-Ritense: Its not fragile at all. This is how changesets work. Nevertheless, its not necessary here. So Ive changed it

changelog/src/main/kotlin/com/ritense/valtimo/changelog/service/ChangelogDeployer.kt
theo-ritense: Shouldn't be an issue when we set ON_DELETE CASCADE at the foreign key to roles.

PR #759, opened 2023-06-19
69226 - Add PBAC support for Note to note module

theo-ritense:
- Added PBAC support for Note to note module
- fixed placeholder matching for user-id and -email


notes/src/main/kotlin/com/ritense/note/service/NoteService.kt
```diff
@@ -24,57 +24,56 @@ import com.ritense.note.domain.Note
...
             AuthorizationRequest(
                 Note::class.java,
-                NoteActionProvider.VIEW
+                NoteActionProvider.LIST_VIEW
                 // TODO: The action type is not actually checked when requesting the spec directly
             ),
             null
         )
-        return noteRepository.findAll(spec.and(SpecificationHelper.byDocumentId(documentId)), pageable)
+        return noteRepository.findAll(spec, pageable)
```
marijnritense, line 58: byDocumentId has to be in, otherwise you'll get results from other documents as well.

marijnritense (approved):

PR #767, opened 2023-06-26
Validate plugin properties after resolving the placeholders.

ThomasMinkeRitense:
Fixed TP #70486: 
As a person implementing the plugin auto configuration
I did use placeholders for values that differ per environment
and expected this to be accepted and the placeholder value be replaced with the environment variable
but experienced an error telling me that it failed parsing the value (in my case as a URI)

### DIFF
--- a/plugin/src/main/kotlin/com/ritense/plugin/service/PluginService.kt
+++ b/plugin/src/main/kotlin/com/ritense/plugin/service/PluginService.kt
@@ -119,12 +119,13 @@ class PluginService(
         ) {
             deletePluginConfiguration(PluginConfigurationId.existingId(deploymentDto.id))
         }
+        resolveProperties(deploymentDto.properties)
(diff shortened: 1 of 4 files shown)


plugin/src/main/kotlin/com/ritense/plugin/service/PluginService.kt
```diff
@@ -119,12 +119,13 @@ class PluginService(
         ) {
             deletePluginConfiguration(PluginConfigurationId.existingId(deploymentDto.id))
         }
+        resolveProperties(deploymentDto.properties)
```
theo-ritense, line 122: resolvesProperties returns a new ObjectNode. It does not change the original deploymentDto.properties.

This change actually breaks the property resolving.

theo-ritense (changes requested):
I don't like that `resolveProperties` does not work immutable. Going to change that and submit it to this PR.

PR #771, opened 2023-07-04
#71162 Retrieving bulk role permissions

Klaas-Ritense:


authorization/src/main/kotlin/com/ritense/authorization/web/rest/PermissionManagementResource.kt
```diff
@@ -0,0 +1,41 @@
...
+import org.springframework.web.bind.annotation.PostMapping
+import org.springframework.web.bind.annotation.RequestBody
+import org.springframework.web.bind.annotation.RequestMapping
+
+@RequestMapping("/api/management", produces = [ValtimoMediaType.APPLICATION_JSON_UTF8_VALUE])
+class PermissionManagementResource(
+    val permissionRepository: PermissionRepository
+) {
+    @PostMapping("/v1/permissions/search")
+    @JsonView(PermissionView.PermissionManagement::class)
```
marijnritense, line 35: I think you need to use a different view for this, otherwise we're leaking additional data with the other endpoints that use this JsonView.

marijnritense (approved):

PR #772, opened 2023-07-04
#66643 Dashboard widget configuration admin endpoints

Klaas-Ritense:


marijnritense (commented):
Can you add tests verifying the endpoints can only be accessed with ROLE_ADMIN?

PR #779, opened 2023-07-11
Feature/37159 upload mime whitelisting

theo-ritense:


form/src/main/java/com/ritense/form/web/rest/impl/FormIoFormFileResource.java
```diff
@@ -28,6 +28,7 @@
 import org.springframework.web.multipart.MultipartFile;
 import org.springframework.web.servlet.view.RedirectView;
 
+@Deprecated(since = "10.7.0")
 public class FormIoFormFileResource implements FormFileResource {
```
Klaas-Ritense, line 32: Isn't this still used for S3?

theo-ritense: Maybe. But I just followed the story on this one.

document-generation/smartdocuments/src/main/kotlin/com/ritense/smartdocuments/client/SmartDocumentsClient.kt
```diff
@@ -67,24 +68,27 @@ class SmartDocumentsClient(
...
-
-        val bodyFlux = webClient().post()
+        val response = webClient().post()
             .uri("/wsxmldeposit/deposit/unattended")
             .contentType(APPLICATION_JSON_UTF8)
             .bodyValue(fixRequest(smartDocumentsRequest))
-            .retrieve()
-            .bodyToFlux<DataBuffer>()
+            .exchange()
+            .block()
```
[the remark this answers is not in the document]
theo-ritense: Good one. The source I got the solution from mentioned it was a solution that was suitable for large files (not kept in memory), but I haven't checked if this is true. Will see if I can add a test (condition) for that.

Klaas-Ritense (approved):

Klaas-Ritense (approved):

PR #780, opened 2023-07-12
#68292 Add PBAC support for tasks in core module

Klaas-Ritense:


core/src/main/kotlin/com/ritense/valtimo/camunda/authorization/CamundaTaskActionProvider.kt
```diff
@@ -0,0 +1,47 @@
...
+    override fun getAvailableActions(): List<Action<CamundaTask>> {
+        return listOf(ASSIGN, ASSIGNABLE, CLAIM, COMPLETE, LIST_VIEW, VIEW)
+    }
+
+    companion object {
+        @JvmField
+        val ASSIGN = Action<CamundaTask>(Action.ASSIGN)
+
+        @JvmField
+        val ASSIGNABLE = Action<CamundaTask>(Action.ASSIGNABLE)
```
theo-ritense, line 33: I still think we should rename the `ASSIGNABLE` action, as it isn't really an action but more of a state. Maybe `LIST_ASSIGNABLE`? Not sure if it should be 'list'. Please make it reflect what action it checks.

core/src/main/kotlin/com/ritense/valtimo/camunda/authorization/CamundaTaskActionProvider.kt
```diff
@@ -0,0 +1,47 @@
...
+
+        @JvmField
+        val CLAIM = Action<CamundaTask>(Action.CLAIM)
+
+        @JvmField
+        val COMPLETE = Action<CamundaTask>(Action.COMPLETE)
+
+        @JvmField
+        val LIST_VIEW = Action<CamundaTask>(Action.LIST_VIEW)
```
theo-ritense, line 42: Same thing. `LIST_VIEW` isn't an action, but more a UI component. `LIST` or `VIEW_LIST` would be an action.

authorization/src/main/kotlin/com/ritense/authorization/AuthorizationRequest.kt
```diff
@@ -19,4 +19,5 @@ package com.ritense.authorization
 interface AuthorizationRequest<T> {
     val resourceType: Class<T>
     val action: Action<T>
+    val user: String?
```
theo-ritense, line 22: I couldn't find any constructor usage where this `user` property is used. Why is it there?

Klaas-Ritense: Its only used [here](https://github.com/valtimo-platform/valtimo-backend-libraries/pull/780/files#diff-e820a081bd1a17e1df5c297fc16e03563c11d8c99e42e62406e43be5f1f43e0aR164).

Its to check this scenario:
When the method `.assign(assignee)` is called, the `assignee` must have the `ASSIGNABLE` permission.

core/src/main/kotlin/com/ritense/valtimo/camunda/authorization/CamundaTaskActionProvider.kt
Klaas-Ritense: These actions already existed. If we want to change that, I would preferably do that in another PR. Because it also changes document permissions

core/src/main/java/com/ritense/valtimo/service/CamundaTaskService.java
```diff
@@ -117,61 +144,109 @@ public CamundaTaskService(
...
-        } catch (AuthorizationException ex) {
-            throw new IllegalStateException("Cannot claim task: the user has no permission.", ex);
-        } catch (ProcessEngineException ex) {
-            throw new IllegalStateException("Cannot claim task: reason is the task doesn't exist.", ex);
+        if (assignee == null) {
+            unassign(taskId);
+        } else {
+            final CamundaTask task = runWithoutAuthorization(() -> findTaskById(taskId));
+            authorizationService.requirePermission(
+                new EntityAuthorizationRequest<>(CamundaTask.class, ASSIGNABLE, assignee, task)
```
marijnritense, line 164: Not sure if this is the way to go. We're not checking permissions for the active user, we're checking permissions for another user. Maybe a separate requirePermissionForUser method, or an EntityAuthorizationForAnotherUserRequest class? Then this would keep the EntityAuthorizationRequest a little bit more clear.

Klaas-Ritense: The `user` field was added to the parent `AuthorizationRequest` because we must always know for which user a authorization is being requested.

But its true that the `user` field will almost always be about the logged-in-user. That's why I created a second constructor. To keep things non-breaking and clean

authorization/src/main/kotlin/com/ritense/authorization/AuthorizationRequest.kt
ivo-ritense: This has been refactored into a specific AuthorizationRequest type when requesting permissions for another user. See DelegateUserEntityAuthorizationRequest

core/src/main/java/com/ritense/valtimo/service/CamundaTaskService.java
ivo-ritense: This has been refactored into a specific AuthorizationRequest type when requesting permissions for another user. See DelegateUserEntityAuthorizationRequest

authorization/src/test/kotlin/com/ritense/authorization/web/rest/PermissionResourceSecurityIT.kt
```diff
@@ -61,7 +61,6 @@ class PermissionResourceSecurityIT : SecuritySpecificEndpointIntegrationTest() {
             )
         )
 
-        permissionRepository.deleteAll()
```
ivo-ritense, line 64: We might want to restore this for saving collections where a permission was deleted.

ivo-ritense (approved):

theo-ritense (approved):

PR #787, opened 2023-07-17
feat: 66985 object-management plugin definition

anna-ritense:
* introduces base plugin definition for a new object-management plugin


zgw/object-management/src/main/kotlin/com/ritense/objectmanagement/ObjectManagementPlugin.kt
```diff
@@ -0,0 +1,74 @@
...
+        TODO("Not implemented!")
+    }
+
+    @PluginAction(
+        key = "delete-object",
+        title = "Delete object",
+        description = "Delete an object by Object Management",
+        activityTypes = [ActivityType.SERVICE_TASK_START]
+    )
+    fun deleteObject(@PluginActionProperty objectUrl: URI): HttpStatus {
```
Klaas-Ritense, line 71: @anna-ritense @ThomasMinkeRitense 
This `delete-object` plugin action already exists in the ObjectenApiPlugin.

Also, if the other 3 plugin actions here would be implemented, they would only need use the ObjectenApiClient. Which means all plugin actions here, could be moved to the ObjectenApiPlugin.

anna-ritense: The idea of this plugin (and it's actions) was to provide a way to use object-management from process-links. We were aware that some actions already exist and are already part of various other plugins. During our refinement we discussed that those previous actions would be deprecated/removed in favour of this uniform way of working with objects.

If this is no longer the desired outcome, then please discuss this feature as a whole with Thomas Minke and let me know so I can stop working on this.

ThomasMinkeRitense: @Klaas-Ritense we had an extensive discussion on this. We can't implement the actions in the objecten API plugin because we'd need a reference to the Object Management Configuration. That would be a circular dependency. That's why we're implementing a new plugin. @anna-ritense Please continue as discussed

Klaas-Ritense: @ThomasMinkeRitense I agree with having a Object-Management-plugin. Just not for these plugins actions. crud actions belong in the objecten-api-plugin

A valid action for Object-Management-plugin would be:
`searchObjects("boom.leeftijd", ">=",  "12")`

anna-ritense: >A valid action for Object-Management-plugin would be:
searchObjects("boom.leeftijd", ">=",  "12")

There is a [story](https://ritense.tpondemand.com/entity/69068-be-object-management-plugin-action-retrieve) for this action, we also need a general get by Id for object-management configuration action. etc.

theo-ritense (approved):

PR #788, opened 2023-07-18
Feature/pbac for non task camunda entities in core

marijnritense:


authorization/src/test/kotlin/com/ritense/authorization/web/rest/RoleManagementResourceSecurityIntTest.kt
```diff
@@ -170,7 +171,8 @@ class RoleManagementResourceSecurityIntTest : SecuritySpecificEndpointIntegratio
     @Test
     @WithMockUser(authorities = [AuthoritiesConstants.ADMIN])
     fun `should have access to update role permissions method with role_admin`() {
-        val basePermission = permissionRepository.findAllByRoleKeyInOrderByRoleKeyAscResourceTypeAsc(listOf("ROLE_USER"))[0]
+        System.out.println(jacksonObjectMapper().writeValueAsString(permissionRepository.findAll()))
```
theo-ritense, line 174: get this out!

document/src/main/java/com/ritense/document/service/impl/JsonSchemaDocumentService.java
```diff
@@ -463,14 +463,36 @@ public void assignUserToDocument(UUID documentId, String assigneeId) {
...
+            try {
+                authorizationService
+                    .requirePermission(
+                        new EntityAuthorizationRequest<>(
+                            JsonSchemaDocument.class,
+                            CLAIM,
+                            document
+                        )
+                    );
+            } catch (Exception e) {
```
theo-ritense, line 476: AccessDeniedException

process-document/src/main/java/com/ritense/processdocument/service/impl/CamundaProcessJsonSchemaDocumentService.java
```diff
@@ -330,11 +332,15 @@ public StartProcessForDocumentResult startProcessForDocument(StartProcessForDocu
     }
 
     public Document getDocument(DelegateExecution execution) {
+        denyAuthorization();
         return getDocument(ProcessInstanceId.fromExecution(execution, CamundaProcessInstanceId.class), execution);
```
theo-ritense, line 336: This method should probably also have a deny?

ivo-ritense (approved):

document/src/main/java/com/ritense/document/service/impl/JsonSchemaDocumentService.java
marijnritense: done!

authorization/src/test/kotlin/com/ritense/authorization/web/rest/RoleManagementResourceSecurityIntTest.kt
marijnritense: yessir

process-document/src/main/java/com/ritense/processdocument/service/impl/CamundaProcessJsonSchemaDocumentService.java
marijnritense: So this particular method should probably be deprecated, as this can only really be called from inside a process, but since this isn't a process bean that's not really possible. I'll add a note about this particular method to the process beans story, so it'll be deprecated then.

PR #798, opened 2023-07-27
Set default of valtimo.camunda.beans-whitelisting to true.

ThomasMinkeRitense:


theo-ritense (approved):

ThomasMinkeRitense:
Draft PR because I suspect tests will fail until TP 72310 has been merged

PR #801, opened 2023-07-28
added Dashboard query widget data endpoint

theo-ritense:


dashboard/src/main/kotlin/com/ritense/dashboard/web/rest/dto/DashboardWidgetDataResultDto.kt
```diff
@@ -0,0 +1,24 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.ritense.dashboard.web.rest.dto
+
+import com.fasterxml.jackson.annotation.JsonUnwrapped
+
+class DashboardWidgetDataResultDto(
```
marijnritense, line 21: Consider making this an abstract class and having several implementations. That way we can still be more specific  with what the result looks like.

dashboard/src/main/kotlin/com/ritense/dashboard/web/rest/dto/DashboardWidgetDataResultDto.kt
```diff
@@ -0,0 +1,24 @@
...
+    val key: String,
+    @JsonUnwrapped val result: Any
```
ivo-ritense, line 23: i'm not sure if i like that this is @JsonUnwrapped. it feels like it would be better to put the object with unknown content in it's own variable.

theo-ritense: I'm was just following the spec. If we want to change that, we need to change the spec.

theo-ritense: changed it.

dashboard/src/main/kotlin/com/ritense/dashboard/web/rest/dto/DashboardWidgetDataResultDto.kt
theo-ritense: As discussed, an abstract doesn't make much sense.

marijnritense (approved):

PR #808, opened 2023-07-31
Fix/sonarqube issues 5

ThomasMinkeRitense:


form-link/src/main/java/com/ritense/formlink/domain/impl/formassociation/CamundaFormAssociation.java
```diff
@@ -20,6 +20,7 @@
 import com.fasterxml.jackson.databind.node.ObjectNode;
 import com.ritense.formlink.domain.FormAssociation;
 import com.ritense.formlink.domain.FormLink;
+import kotlin.jvm.Transient;
```
ivo-ritense, line 23: Is this the class you wanted to apply? Not `javax.persistence.Transient` for example? Not sure what this is supposed to fix.

resource/local-resource/src/main/kotlin/com/ritense/resource/service/LocalResourceService.kt
```diff
@@ -29,62 +29,66 @@ import java.util.UUID
 class LocalResourceService : ResourceService {
 
     override fun store(key: String, multipartFile: MultipartFile): Resource {
-        TODO("Not yet implemented")
+        TODO(NOT_YET_IMPLEMENTED_MSG)
```
marijnritense, line 32: I think for messages like this, we should just mark it as a false positive in sonar. This is how the method was generated. If anything, we should just throw a proper exception as defined by Valtimo.

core/src/main/java/com/ritense/valtimo/web/rest/dto/CustomTaskDto.java
```diff
@@ -29,8 +30,10 @@ public class CustomTaskDto implements Serializable {
 
     private CamundaTaskDto task;
 
+    @Transient
```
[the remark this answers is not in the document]
Cees2505: Sonar has a problem with this fields in a Serialized class  that has no @Transcient or @Serializable annotation. Sonar says that : it could cause program crashes  or is a security risk while opening the door to attackers. No problem here to roll this change back.

form-link/src/main/java/com/ritense/formlink/domain/impl/formassociation/CamundaFormAssociation.java
Cees2505: Sonar has a problem with this fields in a Serialized class  that has no @Transcient or @Serializable annotation. Sonar says that : it could cause program crashes  or is a security risk while opening the door to attackers. No problem here to roll this change back.

core/src/main/java/com/ritense/valtimo/web/rest/dto/CustomTaskDto.java
```diff
@@ -29,8 +30,10 @@ public class CustomTaskDto implements Serializable {
...
     private List<FormField> formFields;
 
+    @Transient
```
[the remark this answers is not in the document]
Cees2505: Sonar has a problem with this fields in a Serialized class that has no @Transcient or @serializable annotation. Sonar says that : it could cause program crashes or is a security risk while opening the door to attackers. No problem here to roll this change back.

PR #809, opened 2023-08-01
Improve-process-link-module

Tom-Ritense:
Added return value for creating/updating processLinks.


form-flow-valtimo/src/main/kotlin/com/ritense/valtimo/formflow/mapper/FormFlowProcessLinkMapper.kt
```diff
@@ -76,7 +75,7 @@ class FormFlowProcessLinkMapper(
             throw RuntimeException("FormFlow definition not found with id ${createRequestDto.formFlowDefinitionId}")
         }
         return FormFlowProcessLink(
-            id = UUID.randomUUID(),
+            id = createRequestDto.processLinkId,
```
theo-ritense, line 78: Because we can now provide an id from the outside, I would expect a check that it is indeed a new processlink and not an existing one.

theo-ritense: It is, but there are separate method for updating and creating process links in the `ProcessLinkService`. We don't want it to be possible to update an existing one using the `createProcessLink` method. The JPA repository uses the `save` method in either case so wouldn't care about an existing entity.

form/src/main/kotlin/com/ritense/form/service/FormSubmissionService.kt
theo-ritense, line 1: This file is missing the copyright header

form-flow-valtimo/src/main/kotlin/com/ritense/valtimo/formflow/web/rest/dto/FormFlowProcessLinkCreateRequestDto.kt
```diff
@@ -20,9 +20,11 @@ import com.fasterxml.jackson.annotation.JsonTypeName
 import com.ritense.processlink.domain.ActivityTypeWithEventName
 import com.ritense.processlink.web.rest.dto.ProcessLinkCreateRequestDto
 import com.ritense.valtimo.formflow.mapper.FormFlowProcessLinkMapper.Companion.PROCESS_LINK_TYPE_FORM_FLOW
+import java.util.UUID
 
 @JsonTypeName(PROCESS_LINK_TYPE_FORM_FLOW)
 data class FormFlowProcessLinkCreateRequestDto(
+    override val processLinkId: UUID = UUID.randomUUID(),
```
ivo-ritense, line 27: This is a breaking change. For kotlin moving this to the end of the properties would avoid breaking, but for Java this will always be a breaking change. Maybe we should add this in major 11?

form-flow-valtimo/src/main/kotlin/com/ritense/valtimo/formflow/mapper/FormFlowProcessLinkMapper.kt
Tom-Ritense: Will revert to having a return var instead of provided id.

Klaas-Ritense (approved):

PR #815, opened 2023-08-02
Fix/sonarqube issues 6

ThomasMinkeRitense:


ThomasMinkeRitense:
Decision: won't fix the generics-related warnings.

PR #816, opened 2023-08-03
Bugfix/loading process variables for summary form

marijnritense:


form-link/src/main/java/com/ritense/formlink/service/impl/CamundaFormAssociationService.java
```diff
@@ -343,7 +343,7 @@ private void prefillForm(
         final ObjectNode extendedDocumentContent = (ObjectNode) document.content().asJson();
         extendedDocumentContent.set("metadata", buildMetaDataObject(document));
 
-        if (taskInstanceId.isEmpty()) {
+        if (taskInstanceId.isEmpty() && !formDefinition.getName().endsWith(".summary")) {
```
theo-ritense, line 346: It really feels like a workaround having to exclude ".summary" definitions. Then again: the FormAssociation is about to be deprecated/removed, so I'm torn on whether to allow it.

marijnritense: Process variables aren't supposed to be used in `.summary`. And yes, this will be removed at some point. Not sure what there is to be thorn about.

theo-ritense (approved):

PR #822, opened 2023-08-04
76423- Fixed Widget Configuration order on edit

theo-ritense:


ivo-ritense (approved):
You could make a test for this

PR #839, opened 2023-08-18
Feature/77093 pbac support document definitions

theo-ritense:


case/src/main/kotlin/com/ritense/case/web/rest/CaseDefinitionResource.kt
ivo-ritense, line 1: Shouldn't this API be /api/management/v1/?

document/src/main/java/com/ritense/document/web/rest/DocumentDefinitionResource.java
```diff
@@ -44,32 +44,30 @@
...
+    @PostMapping(value = "/v1/document-definition", consumes = APPLICATION_JSON_VALUE)
     ResponseEntity<DeployDocumentDefinitionResult> deployDocumentDefinition(
         @Valid @RequestBody DocumentDefinitionCreateRequest request
     );
 
-    @DeleteMapping("/document-definition/{name}")
+    @DeleteMapping("/v1/document-definition/{name}")
     ResponseEntity<UndeployDocumentDefinitionResult> removeDocumentDefinition(@PathVariable String name);
-
-    @GetMapping("/document-definition/{documentDefinitionName}/roles")
```
Klaas-Ritense, line 70: This will require some FE changes, right?

case/src/main/kotlin/com/ritense/case/web/rest/CaseDefinitionResource.kt
Klaas-Ritense: I think this should be done in a different story. Since there are other endpoint that also miss the `/management` part. Ok?

Klaas-Ritense (approved):

document/src/main/java/com/ritense/document/web/rest/DocumentDefinitionResource.java
Klaas-Ritense: See: https://github.com/valtimo-platform/valtimo-frontend-libraries/pull/552

PR #851, opened 2023-08-29
Feature/24771 CVE scanning

RemcoRitense:
Added java dependency vulnerability scan.


.java-version
Hildebrand-Ritense, line 1: Personally don't like this. It complicates version matching with what version you have specifically running locally. Also, this is out of scope for this PR. I do know others like it, because it's a more precise match, but I've never seen bugs coming from a specific Java minor. I'm going to leave this to the Valtimo team...

gradle/build.gradle
Hildebrand-Ritense, line 1: in what scenario did you need this? I don't think this was needed for the other (license-report) module.

gradle/cve-report/cve-exclusions.xml
Hildebrand-Ritense, line 1: Can you explain why we're excluding guava? Good to have this documented :)

Hildebrand-Ritense, line 1: let's rework the contents a bit so it's clear it's an example / placeholder and that it needs the configuration in the gradle/cve-report/build.gradle file for it to do anything.

Hildebrand-Ritense (commented):
almost there 👍

Hildebrand-Ritense (approved):
lgtm

gradle/build.gradle
```diff
@@ -0,0 +1,6 @@
+tasks.withType(PublishToMavenRepository) {
+    enabled = false
+}
+tasks.withType(PublishToMavenLocal) {
```
marijnritense, line 4: Why is this being disabled?

RemcoRitense: Let me look into this, this piece of code was found in multiple locations and I might have missed this one.

Hildebrand-Ritense: I think I added this way back because the gradle folder, being a folder/module with submodules, shouldn't itself be published. I asked Remco to double check this (see my comment 2 days ago) since this was added way back.

marijnritense: Okay, that makes sense!

PR #861, opened 2023-09-01
#75225 Remove roles (entitlements)

Klaas-Ritense:


core/src/main/resources/config/liquibase/changelog/initial-domain-tables.xml
```diff
@@ -20,25 +20,6 @@
     xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
     xsi:schemaLocation="http://www.liquibase.org/xml/ns/dbchangelog http://www.liquibase.org/xml/ns/dbchangelog/dbchangelog-3.4.xsd">
 
-    <changeSet id="0" author="jhipster">
```
marijnritense, line 23: I'm not entirely sure about just removing every changelog, won't this mean old tables will not be cleaned up for existing applications?

marijnritense (approved):

marijnritense (approved):

PR #862, opened 2023-09-01
#77111 Fix: PBAC 'contains' operator

Klaas-Ritense:


authorization/src/main/kotlin/com/ritense/authorization/permission/condition/ExpressionPermissionCondition.kt
```diff
@@ -53,11 +53,15 @@ data class ExpressionPermissionCondition<V : Comparable<V>>(
         }
 
         val pathValue = try {
-            JsonPath.read<V?>(jsonValue, path)
+            JsonPath.read<Any?>(jsonValue, path)
```
marijnritense, line 56: Can we instead check for V first, and if not V then Collection<V>?

Klaas-Ritense: Not sure what you mean but the type checking is done below

marijnritense: Right, and we can't check this first because we don't know what kind of value we're trying to read. Okay, that makes sense. Not entirely happy with `Collection<*>` though. Ideally we'd check if this was a `Collection<clazz>`

PR #872, opened 2023-09-06
#50055 Rename form flow API endpoints

Klaas-Ritense:


form-flow-valtimo/src/main/kotlin/com/ritense/valtimo/formflow/web/rest/FormFlowResource.kt
```diff
@@ -57,7 +57,7 @@ class FormFlowResource(
         return ResponseEntity.ok(GetFormFlowStateResult(instance.id.id, openStep(stepInstance)))
     }
 
-    @PostMapping("/v1/form-flow/{formFlowId}/step/{stepInstanceId}")
+    @PostMapping(value = ["/v1/form-flow/instance/{formFlowId}/step/instance/{stepInstanceId}","/v1/form-flow/{formFlowId}/step/{stepInstanceId}"]) // TODO: deprecate "/v1/form-flow/{formFlowId}/step/{stepInstanceId}"
```
marijnritense, line 60: So there's still some TODOs in here, can you do those? I think they're a part of this story right?

theo-ritense (commented):
Do we really want separate methods for the new endpoints? The deprecation annotation only adds value when it's being called from implementation code, which will never be the case.

I think we could get away with just adding another path to the mapping annotation, and a deprecation-comment on the same line.

ivo-ritense (approved):

PR #875, opened 2023-09-08
#75275 Add logging on PBAC data filtering

Klaas-Ritense:
And decrease the amount of logging by the /open/count endpoint.

### DIFF
--- a/authorization/src/main/kotlin/com/ritense/authorization/ValtimoAuthorizationService.kt
+++ b/authorization/src/main/kotlin/com/ritense/authorization/ValtimoAuthorizationService.kt
@@ -35,15 +35,12 @@ class ValtimoAuthorizationService(
     private val actionProviders: List<ResourceActionProvider<*>>,
     private val permissionRepository: PermissionRepository,
     private val userManagementService: UserManagementService
-): AuthorizationService {
+) : AuthorizationService {
     override fun <T : Any> requirePermission(
         request: EntityAuthorizationRequest<T>
     ) {
         if (!hasPermission(request)) {
-            if (request.action.key == Action.DENY) {
-                logger.error { "Access denied on '${request.resourceType}'. This generally indicates attempting to " +
-                    "access a resource without considering authorization. Please refer to the Valtimo documentation." }
-            } else {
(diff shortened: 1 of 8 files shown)


theo-ritense (commented):
I don't think this is what is described in the story. Instead of the applied permissions, it logging every permission.

authorization/src/main/kotlin/com/ritense/authorization/ValtimoAuthorizationService.kt
```diff
@@ -116,6 +124,21 @@ class ValtimoAuthorizationService(
...
     }
 
+    private fun logPermissions(request: AuthorizationRequest<*>, permissions: List<Permission>? = null) {
+        if (request.action.key == Action.DENY) {
+            logger.error {
+                "Access denied on '${request.resourceType}'. This generally indicates attempting to " +
+                    "access a resource without considering authorization. Please refer to the Valtimo documentation."
+            }
+        } else {
+            val userPermissions = permissions ?: getPermissions(request)
```
Klaas-Ritense, line 134: @theo-ritense Its only logging the applied permissions. But it does log a lot. For example when assigning a 'developer' to a task, it logs:
```
User 'admin@example.com' is requesting permission to 'view:CamundaTask' and has the following matching permissions: [b1108a2d-66b0-4abf-bcd3-76760c6dafbd:ROLE_USER]
User 'admin@example.com' is requesting permission to 'view:CamundaTask' and has the following matching permissions: [b1108a2d-66b0-4abf-bcd3-76760c6dafbd:ROLE_USER]
User 'admin@example.com' is requesting permission to 'assign:CamundaTask' and has the following matching permissions: [16713494-e9f4-4c25-9bdf-f4c93030ef9d:ROLE_ADMIN]
User 'developer@example.com' is requesting permission to 'assignable:CamundaTask' and has the following matching permissions: [02fb3459-6248-4b5b-a72b-57c1cb7a2594:ROLE_USER]
```

theo-ritense (dismissed):
I don't think this is what is described in the story. Instead of the applied permissions, it logging every permission.

theo-ritense (approved):

PR #896, opened 2023-09-19
#77111 support contains in authorization permission conditions

Klaas-Ritense:


authorization/src/main/kotlin/com/ritense/authorization/ValtimoAuthorizationService.kt
```diff
@@ -128,16 +128,18 @@ class ValtimoAuthorizationService(
...
+        if (!AuthorizationContext.ignoreAuthorization) {
+            if (request.action.key == Action.DENY) {
+                logger.error {
+                    "Access denied on '${request.resourceType}'. This generally indicates attempting to " +
+                        "access a resource without considering authorization. Please refer to the Valtimo documentation."
+                }
+            } else {
+                val permissionsLogLine = permissions.joinToString(", ") { "${it.id}:${it.role.key}" }
+                val logLine =
+                    "Requesting permissions '${request.action.key}:${request.resourceType.simpleName}' for user '${request.user}' and found matching permissions: [$permissionsLogLine]"
```
marijnritense, line 140: This is a bit misleading because we're ignoring authorization altogether. Maybe we should have the log line reflect that.

Klaas-Ritense: Added a logline for that

marijnritense (approved):

PR #898, opened 2023-09-22
Fix pbac document:create permission

Klaas-Ritense:
Problem:
Creating a case was possible without permissions


theo-ritense (commented):
From the PR, I can't see any CREATE check being added. Where is it checked exactly?

form/src/main/kotlin/com/ritense/form/service/impl/DefaultFormSubmissionService.kt
```diff
@@ -98,33 +98,31 @@ open class DefaultFormSubmissionService(
                 )
             }
 
-            return AuthorizationContext.runWithoutAuthorization {
```
Klaas-Ritense, line 101: @theo-ritense The `CREATE` check already exists but is ignored because of this `runWithoutAuthorization` that wraps everything

ivo-ritense (approved):

ivo-ritense:
You might want to write a test to check this specifically

PR #927, opened 2023-09-27
#78212 Case tab Autodeployment

Klaas-Ritense:


case/src/main/kotlin/com/ritense/case/deployment/CaseTabChangeset.kt
```diff
@@ -0,0 +1,31 @@
...
+import com.fasterxml.jackson.annotation.JsonProperty
+import com.ritense.case.web.rest.dto.CaseTabDto
+
+class CaseTabChangeset(
+    val changesetId: String,
+    @JsonProperty("case-definitions")
+    val caseDefinitions: List<CaseDefinitionsTabCollection>
+)
+
+class CaseDefinitionsTabCollection(
```
theo-ritense, line 28: I think we should define public classes in their own file

case/src/main/kotlin/com/ritense/case/deployment/CaseTabChangeset.kt
```diff
@@ -0,0 +1,31 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.ritense.case.deployment
+
+import com.fasterxml.jackson.annotation.JsonProperty
+import com.ritense.case.web.rest.dto.CaseTabDto
+
+class CaseTabChangeset(
```
theo-ritense, line 22: data class?

case/src/main/kotlin/com/ritense/case/deployment/CaseTabChangeset.kt
theo-ritense, line 28: data class?

case/src/main/kotlin/com/ritense/case/deployment/CaseTabDeployer.kt
```diff
@@ -0,0 +1,75 @@
...
+            ChangesetDetails(
+                changesetId = changeset.changesetId,
+                valueToChecksum = changeset.caseDefinitions,
+                key = KEY,
+                deploy = { deploy(changeset.caseDefinitions) }
+            )
+        )
+    }
+
+    fun deploy(caseDefinitions: List<CaseDefinitionsTabCollection>) {
```
theo-ritense: private ?

case/src/main/kotlin/com/ritense/case/domain/CaseTab.kt
```diff
@@ -0,0 +1,40 @@
...
+    val id: CaseTabId,
+
+    val name: String,
+
+    val tabOrder: Int,
+
+    @Enumerated(EnumType.STRING)
+    val type: CaseTabType,
+
+    val content: String,
```
theo-ritense: Can we have a property name that better describes what's in there? From the examples, it looks more like a component-key or something

case/src/main/kotlin/com/ritense/case/web/rest/dto/CaseTabDto.kt
```diff
@@ -0,0 +1,26 @@
...
+
+package com.ritense.case.web.rest.dto
+
+import com.ritense.case.domain.CaseTabType
+
+data class CaseTabDto(
+    val key: String,
+    val name: String,
+    val type: CaseTabType,
+    val content: String,
```
theo-ritense: Please see my comment on the entity class and change accordingly if needed.

case/src/test/kotlin/com/ritense/case/deployment/CaseTabDeployerIT.kt
```diff
@@ -0,0 +1,103 @@
...
+import org.junit.jupiter.api.Test
+import org.mockito.kotlin.whenever
+import org.springframework.beans.factory.annotation.Autowired
+import org.springframework.data.domain.Sort
+import org.springframework.transaction.annotation.Transactional
+import java.time.Instant
+
+@Transactional
+class CaseTabDeployerIT : BaseIntegrationTest() {
+    @Autowired
```
theo-ritense: For integration tests, it makes more sense to autowire these beans in the constructor

case/src/main/kotlin/com/ritense/case/domain/CaseTab.kt
ivo-ritense: The issue is that this can be multipole things depending on the type. It's either the form key, system tab name or angular component name.

case/src/main/kotlin/com/ritense/case/deployment/CaseTabChangeset.kt
ivo-ritense: What would be the use of that?

case/src/main/kotlin/com/ritense/case/deployment/CaseTabChangeset.kt
ivo-ritense: This could be package private i think. The only use of these classes is to represent the structure of the import file. I think having extra class files would only make the project structure less clear.

case/src/test/kotlin/com/ritense/case/deployment/CaseTabDeployerIT.kt
Klaas-Ritense: Changed

case/src/main/kotlin/com/ritense/case/deployment/CaseTabDeployer.kt
Klaas-Ritense: Changed

case/src/main/kotlin/com/ritense/case/domain/CaseTab.kt
theo-ritense: `contentKey` then? `content` gives me vibes of a JSON-object, which is not what seems to be intended.

case/src/main/kotlin/com/ritense/case/deployment/CaseTabChangeset.kt
theo-ritense: It generates the `toString`, `copy`, `equals` and `hashCode` methods for you.
But if you really don't need it, that's fine with me.

ivo-ritense (approved):

PR #940, opened 2023-09-29
#81711 create tab list endpoints

Klaas-Ritense:


app/gzac/src/main/resources/config/case-tabs/bezwaar.case-tabs.json
```diff
@@ -0,0 +1,52 @@
...
+    "changesetId": "bezwaar-tabs-v1",
+    "case-definitions": [
+        {
+            "key": "bezwaar",
+            "tabs": [
+                {
+                    "name": "Summary",
+                    "key": "summary",
+                    "type": "standard",
+                    "content": "bezwaar.summary"
```
ivo-ritense: i think the content here should just be "summary" and the way it will find the right form is the same as before. to use other forms to display data can be done using the formio type.

Klaas-Ritense: Changed

ivo-ritense (approved):

theo-ritense (approved):

PR #944, opened 2023-10-02
#78206 Change tabMan order of tabs

Klaas-Ritense:


case/src/main/kotlin/com/ritense/case/web/rest/CaseTabManagementResource.kt
```diff
@@ -30,12 +32,12 @@ import org.springframework.web.bind.annotation.RequestBody
...
 
 @RequestMapping("/api/management", produces = [APPLICATION_JSON_UTF8_VALUE])
-class CaseTabManagementResource(
+open class CaseTabManagementResource(
     private val caseTabService: CaseTabService
 ) {
 
     @PostMapping("/v1/case-definition/{caseDefinitionName}/tab")
-    fun createCaseTab(
+    open fun createCaseTab(
```
theo-ritense, line 40: functions don't need to be specified `open` when the class is already `open`, right?

case/src/main/kotlin/com/ritense/case/web/rest/dto/CaseTabUpdateOrderDto.kt
```diff
@@ -0,0 +1,26 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.ritense.case.web.rest.dto
+
+import com.ritense.case.domain.CaseTabType
+
+data class CaseTabUpdateOrderDto(
```
theo-ritense, line 21: This is the same as `CaseTabUpdateDto` if I remember correctly. Can't we use the same one?

case/src/main/kotlin/com/ritense/case/web/rest/CaseTabManagementResource.kt
Klaas-Ritense: Unfortunately, if even a single function here is not `open`, I get an error.

All the `open` in this class are needed to make sure the one `@RunWithoutAuthorization` works.

case/src/main/kotlin/com/ritense/case/web/rest/dto/CaseTabUpdateOrderDto.kt
Klaas-Ritense: There is a small difference. This dto also includes a `key` field

theo-ritense (approved):

PR #947, opened 2023-10-03
Added relatedFiles.size to 'case' value resolver.

ThomasMinkeRitense:
Enables implementations to show the number of related files in the case list.


process-document/src/main/kotlin/com/ritense/processdocument/resolver/DocumentTableValueResolver.kt
```diff
@@ -80,6 +80,7 @@ class DocumentTableValueResolver(
                 "assigneeFullName" -> document.assigneeFullName()
                 "version" -> document.version()
                 "sequence" -> document.sequence()
+                "relatedFiles.size" -> document.relatedFiles().size
```
ivo-ritense, line 83: You are aware this doesn't work with the documenten api?

PR #956, opened 2023-10-06
fix: plugins configuration export order

Klaas-Ritense:
The first plugin in the export file gets created first. If that plugin happens to depend on another plugin, an error is thrown.

### DIFF
--- a/plugin/src/main/kotlin/com/ritense/plugin/web/rest/PluginConfigurationResource.kt
+++ b/plugin/src/main/kotlin/com/ritense/plugin/web/rest/PluginConfigurationResource.kt
@@ -17,6 +17,7 @@
 package com.ritense.plugin.web.rest
 
 import com.ritense.plugin.domain.ActivityType
+import com.ritense.plugin.domain.PluginConfiguration
 import com.ritense.plugin.domain.PluginConfigurationId
 import com.ritense.plugin.service.PluginConfigurationSearchParameters
 import com.ritense.plugin.service.PluginService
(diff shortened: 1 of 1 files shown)


plugin/src/main/kotlin/com/ritense/plugin/web/rest/PluginConfigurationResource.kt
```diff
@@ -98,6 +100,11 @@ class PluginConfigurationResource(
     @GetMapping("/v1/plugin/configuration/export")
     fun exportPluginConfiguration(): ResponseEntity<List<PluginConfigurationExportDto>> {
         val pluginConfigurations = pluginService.getPluginConfigurations(PluginConfigurationSearchParameters())
+            .sortedWith(comparingInt<PluginConfiguration> { pluginConfiguration ->
+                pluginConfiguration.properties?.fieldNames()?.asSequence()?.count { it.contains("PluginConfiguration") } ?: 0
```
theo-ritense, line 104: What if we have triple-nested configurations? I don't think you can get away with just sorting then. You'd have to follow the references (build a tree) somehow.

Klaas-Ritense: Its true that this approach wont work for future plugins. But it works for all current plugins. imo its good enough for a hidden endpoint. 

But if you're in need of a challenge, you can give it a shot :)

theo-ritense: Challenge accepted! Still going to accept the change for now though.

theo-ritense (approved):

PR #965, opened 2023-10-09
Added PBAC support for case tabs

ivo-ritense:


case/src/main/kotlin/com/ritense/case/repository/CaseTabSpecification.kt
```diff
@@ -0,0 +1,73 @@
...
+        query: CriteriaQuery<*>,
+        criteriaBuilder: CriteriaBuilder
+    ): Predicate {
+        // Filter the permissions for the relevant ones and use those to  find the filters that are required
+        // Turn those filters into predicates
+        val groupList = query.groupList.toMutableList()
+        groupList.add(root.get<UUID>("id"))
+        query.groupBy(groupList)
+
+        val predicates = permissions.stream()
```
theo-ritense, line 50: This piece of code seems to be repeating itself for every specification that we make. Can we deduplicate this somehow?

Klaas-Ritense (dismissed):

objects-api/src/main/kotlin/com/ritense/objectsapi/productaanvraag/ProductAanvraagListener.kt
```diff
@@ -48,8 +49,9 @@ class ProductAanvraagListener(
                     val productAanvraag = connector.getProductAanvraag(productAanvraagId)
                     val typeMapping = connector.getTypeMapping(productAanvraag.type)
                     val aanvragerRolTypeUrl = connector.getAanvragerRolTypeUrl()
-
-                    productAanvraagService.createDossier(productAanvraag, typeMapping, aanvragerRolTypeUrl)
+                    AuthorizationContext.runWithoutAuthorization {
+                        productAanvraagService.createDossier(productAanvraag, typeMapping, aanvragerRolTypeUrl)
+                    }
```
Klaas-Ritense, line 54: Hmm this should also go to 11.0.0

Klaas-Ritense (approved):

case/src/main/kotlin/com/ritense/case/repository/CaseTabSpecification.kt
ivo-ritense: It seems hard to do because of the filter on class type. It might be possible with some reflection. I'll take a look outside the scope of this story.

theo-ritense: It could be sufficient to split it into some abstract method that accepts the class as an argument?

PR #966, opened 2023-10-10
#71947 Add System.getProperty() support to plugin autodeployment

Klaas-Ritense:


plugin/src/test/kotlin/com/ritense/plugin/service/PluginServiceIT.kt
```diff
@@ -343,4 +345,37 @@ internal class PluginServiceIT : BaseIntegrationTest() {
...
         }
     }
+
+    @Test
+    @Transactional
+    fun `should deploy plugin with resolved properties`() {
+        val pluginConfigurationId = UUID.randomUUID()
+        System.setProperty("MY_URL_PLACEHOLDER", "https://www.example.com/")
+        System.setProperty("MY_FIRST_PLACEHOLDER", "first_value")
+        System.setProperty("MY_SECOND_PLACEHOLDER", "second_value")
```
theo-ritense, line 355: Could you tests an `env` value as well?

Klaas-Ritense: I've tried. Unfortunately, its very difficult to set an environment variable. Since it works different for every system.

theo-ritense (approved):

PR #967, opened 2023-10-11
fix case-tabs

Klaas-Ritense:


app/gzac/build.gradle
```diff
@@ -85,7 +85,7 @@ bootRun {
             f.readLines().each() {
                 if (!it.isEmpty() && !it.startsWith("#")) {
                     def (key, value) = it.tokenize('=')
-                    environment key.trim(), value.trim()
```
marijnritense, line 88: Why no more trim?

Klaas-Ritense: trim() will throw an error when the .env.properties contains an empty value like: `SOME_VAR=`. Because the `value` is then null

theo-ritense: The right way would be to add a null check.

theo-ritense: @marijnritense could you check again?

marijnritense (approved):

PR #970, opened 2023-10-17
#83476 Support targetKey for saving form values with the value-resolvers

Klaas-Ritense:


form/src/main/kotlin/com/ritense/form/service/impl/DefaultFormSubmissionService.kt
```diff
@@ -209,16 +209,20 @@ open class DefaultFormSubmissionService(
...
     ): Pair<String, Any>? {
-        return FormIoFormDefinition.GET_SOURCE_KEY.apply(field).getOrNull()?.let { sourceKey ->
-            FormIoFormDefinition.GET_KEY.apply(field).getOrNull()?.let { inputKey ->
-                convertNodeValue(formData.at("/$inputKey"))
-            }?.let { value ->
-                Pair(sourceKey, value)
-            }
+        return FormIoFormDefinition.getTargetKey(field).getOrNull()?.let { targetKey ->
+            getFormValue(field, formData)?.let { value -> Pair(targetKey, value) }
+        } ?: FormIoFormDefinition.getSourceKey(field).getOrNull()?.let { sourceKey ->
```
marijnritense, line 218: Is there a test that covers this scenario? I don't think I see it

Klaas-Ritense: Yes there is. The 'form-example-various-prefill-fields.json' has a `dateOfBirth` with a sourceKey but no targetKey

marijnritense (approved):

PR #974, opened 2023-10-18
Added zaakurl and due date to portaltask object

ivo-ritense:


zgw/portaaltaak/src/main/kotlin/com/ritense/portaaltaak/PortaaltaakPlugin.kt
```diff
@@ -91,6 +93,7 @@ class PortaaltaakPlugin(
         @PluginActionProperty receiver: TaakReceiver,
         @PluginActionProperty identificationKey: String?,
         @PluginActionProperty identificationValue: String?,
+        @PluginActionProperty verloopDurationInDays: Long?,
```
Klaas-Ritense, line 96: This doesn't break when an implementation already has this plugin-action configured?

ivo-ritense: The tests already covered a null argument situation. I added another test to verify that the framework will pass in a null if the property is missing from the configuration. So this shouldn't break.

Klaas-Ritense (dismissed):

Klaas-Ritense (approved):

PR #976, opened 2023-10-19
fix: dmn uploads in 10.8

mbritense:
See comment https://ritense.tpondemand.com/entity/71945-be-cant-upload-dmn


ivo-ritense (approved):
good job!

PR #979, opened 2023-10-20
fixed NOT_EQUAL_TO and EQUAL_TO operator for PBAC

theo-ritense:


ivo-ritense (dismissed):

ivo-ritense:
You could add a test to verify it works like that.

marijnritense (approved):

PR #981, opened 2023-10-23
#84725 outbox module

Klaas-Ritense:


outbox/src/main/kotlin/com/ritense/outbox/OutboxAutoConfiguration.kt
```diff
@@ -0,0 +1,102 @@
...
+        OutboxMessageRepository::class
+    ]
+)
+@EntityScan(basePackages = ["com.ritense.outbox"])
+@AutoConfigureAfter(DataSourceAutoConfiguration::class, HibernateJpaAutoConfiguration::class)
+@EnableConfigurationProperties(LiquibaseProperties::class)
+class OutboxAutoConfiguration {
+
+    @Bean
+    @ConditionalOnMissingBean(OutboxLiquibaseRunner::class)
```
theo-ritense, line 49: This should also be conditional on class `LiquibaseMasterChangeLogLocation`. If that class is on the classpath, that should be used. The dependency on `contract` should be `provided`

outbox/src/main/kotlin/com/ritense/outbox/OutboxAutoConfiguration.kt
```diff
@@ -0,0 +1,102 @@
...
+
+    @Bean
+    @ConditionalOnMissingBean(PollingPublisherJob::class)
+    fun pollingPublisherJob(
+        pollingPublisherService: PollingPublisherService
+    ): PollingPublisherJob {
+        return PollingPublisherJob(pollingPublisherService)
+    }
+
+    @Order(LOWEST_PRECEDENCE)
```
theo-ritense: Why the lowest?

outbox/src/main/kotlin/com/ritense/outbox/OutboxMessage.kt
```diff
@@ -0,0 +1,45 @@
...
+import javax.persistence.Entity
+import javax.persistence.Id
+import javax.persistence.Table
+
+@Entity
+@Table(name = "outbox_message")
+class OutboxMessage(
+
+    @Id
+    @Column(name = "id", columnDefinition = "BINARY(16)", updatable = false)
```
theo-ritense: I don't think the columnDefinition should be here, as that depends on the underlying dbms

outbox/src/main/kotlin/com/ritense/outbox/OutboxService.kt
```diff
@@ -0,0 +1,65 @@
...
+import org.springframework.transaction.annotation.Transactional
+import java.util.UUID
+
+open class OutboxService(
+    private val outboxMessageRepository: OutboxMessageRepository,
+    private val objectMapper: ObjectMapper,
+) {
+
+    @Transactional(propagation = Propagation.MANDATORY)
+    open fun send(message: Any) {
```
theo-ritense, line 31: I don't believe we should support sending a message without providing a specific type. Moving classes will affect the type.

outbox/src/main/kotlin/com/ritense/outbox/publisher/PollingPublisherJob.kt
```diff
@@ -0,0 +1,29 @@
...
+
+package com.ritense.outbox.publisher
+
+import org.springframework.scheduling.annotation.Scheduled
+
+class PollingPublisherJob(
+    private val pollingPublisherService: PollingPublisherService
+) {
+
+    @Scheduled(cron = "\${valtimo.outbox.pollingpublisher.job:*/10 * * * * *}")
```
theo-ritense: I prefer using `fixedRateString` here, because it really a rate you'd want to specify here, and not 'once every sunday in an odd week in may'

outbox/src/main/resources/config/liquibase/changelog/20231020-add-outbox-table.xml
```diff
@@ -0,0 +1,39 @@
...
+            <column name="id" type="${uuidType}">
+                <constraints nullable="false" primaryKey="true" primaryKeyName="outboxPK"/>
+            </column>
+            <column name="message" type="${jsonType}">
+                <constraints nullable="false"/>
+            </column>
+            <column name="event_type" type="VARCHAR(1024)">
+                <constraints nullable="false"/>
+            </column>
+            <column name="created_on" type="DATETIME">
```
theo-ritense, line 30: Since we're sorting on this date, I'd expect an index on it. So either we add the index, or we don't sort.

outbox/src/main/kotlin/com/ritense/outbox/OutboxAutoConfiguration.kt
Klaas-Ritense: There's no need for a dependency on `contract`. Liquibase is ran separately

outbox/src/main/kotlin/com/ritense/outbox/OutboxAutoConfiguration.kt
Klaas-Ritense: So that any other MessagePublisher will get a higher priority

outbox/src/main/kotlin/com/ritense/outbox/OutboxService.kt
Klaas-Ritense: Why not? The function is here because 9 out of 10 times the message type will be the class name. 

Also, moving classes won't affect the type

Klaas-Ritense: After looking into it, the eventType won't be needed in the outbox table. The eventType will be inside the BaseEvent which will be saved inside the outbox.message JSON

outbox/src/main/kotlin/com/ritense/outbox/OutboxAutoConfiguration.kt
ivo-ritense: I doesn't look like we support more than one instance though.

outbox/src/main/kotlin/com/ritense/outbox/OutboxService.kt
ivo-ritense: I feel like that type should be a base event in stead of any though. We want to guarenteee that there are a few properties in the message that is sent so that consumers can filter based on it. Issue is that the event is only introduced in the next PR.

Klaas-Ritense: Yes, the base-event is introduced in the next PR. In that PR the `Any` here is changed to the base event

outbox/src/main/kotlin/com/ritense/outbox/publisher/LoggingMessagePublisher.kt
```diff
@@ -0,0 +1,35 @@
...
+ * limitations under the License.
+ */
+
+package com.ritense.outbox.publisher
+
+import com.fasterxml.jackson.databind.ObjectMapper
+import com.ritense.outbox.OutboxMessage
+import mu.KotlinLogging
+
+open class LoggingMessagePublisher(
```
ivo-ritense: Please add a TODO saying this is temporary

outbox/src/main/kotlin/com/ritense/outbox/OutboxMessageSaved.kt
```diff
@@ -0,0 +1,21 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" basis,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.ritense.outbox
+
+data class OutboxMessageSaved(
```
ivo-ritense: Delete this please.

outbox/src/main/kotlin/com/ritense/outbox/publisher/PollingPublisherService.kt
```diff
@@ -0,0 +1,61 @@
...
+     * Poll messages from the outbox table and publishes them in the correct order.
+     */
+    open fun pollAndPublishAll() {
+        if (polling.compareAndSet(false, true)) {
+            do {
+                try {
+                    TransactionTemplate(platformTransactionManager).executeWithoutResult {
+                        val oldestMessage = outboxService.getOldestMessage()
+                        if (oldestMessage != null) {
+                            logger.info("Sending message '${oldestMessage.id}': ${oldestMessage.message}")
```
ivo-ritense: Change this to debug and remove the full message form the string.

outbox/src/test/kotlin/com/ritense/outbox/OutboxServiceIntTest.kt
```diff
@@ -0,0 +1,63 @@
...
+package com.ritense.outbox
+
+import com.fasterxml.jackson.databind.ObjectMapper
+import org.assertj.core.api.Assertions.assertThat
+import org.junit.jupiter.api.Test
+import org.junit.jupiter.api.assertThrows
+import org.springframework.beans.factory.annotation.Autowired
+import org.springframework.transaction.annotation.Transactional
+
+class OutboxServiceIntTest : BaseIntegrationTest() {
```
ivo-ritense, line 26: Can we add a test that proves the skip locked thing works?

outbox/src/test/kotlin/com/ritense/outbox/publisher/PollingPublisherServiceIntTest.kt
```diff
@@ -0,0 +1,52 @@
...
+import com.fasterxml.jackson.databind.ObjectMapper
+import com.ritense.outbox.BaseIntegrationTest
+import com.ritense.outbox.OutboxMessage
+import com.ritense.outbox.OutboxMessageRepository
+import org.assertj.core.api.Assertions.assertThat
+import org.junit.jupiter.api.Test
+import org.springframework.beans.factory.annotation.Autowired
+import org.springframework.transaction.annotation.Transactional
+
+class PollingPublisherServiceIntTest : BaseIntegrationTest() {
```
ivo-ritense, line 32: Can we add a test that proves the AtomicBoolean locking works the way it should?

ivo-ritense (approved):

PR #985, opened 2023-10-24
feat:  create base event implementation

mbritense:
PR merges into outbox PR because it is based on that. Closes https://ritense.tpondemand.com/entity/84759-be-create-base-event-outbox-service


outbox/build.gradle
```diff
@@ -22,6 +22,7 @@ dockerCompose {
 }
 
 dependencies {
+    implementation project(':contract')
```
ivo-ritense: I thought the outbox module should not have dependencies on other valtimo modules. Is this required?

outbox/src/main/kotlin/com/ritense/outbox/domain/BaseEvent.kt
Klaas-Ritense, line 1: Missing copyright header

outbox/build.gradle
mbritense: Yes, it is required for `val userId: String = SecurityUtils.getCurrentUserLogin() ?: "System",
    val roles: String = SecurityUtils.getCurrentUserRoles().joinToString(),`

ivo-ritense: We want to remove this dependency. Instead we want a bean that has a copy of these methods. This bean can be overridden in implementations if teams want to use a different spring authentication. `UserProvider` This needs to be an open class and the bean registration needs to use ConditionalOnMissingBean

outbox/src/main/kotlin/com/ritense/outbox/domain/BaseEvent.kt
```diff
@@ -0,0 +1,34 @@
...
+import com.ritense.valtimo.contract.utils.SecurityUtils
+import java.time.LocalDateTime
+import java.util.*
+
+abstract class BaseEvent (
+    open val id: UUID  = UUID.randomUUID(),
+    open var source: String?,
+    open val specversion: String,
+    open val type: String,
+    open val data: LocalDateTime = LocalDateTime.now(),
```
ivo-ritense: Please change the variable name to `date` instead.

outbox/src/main/kotlin/com/ritense/outbox/domain/BaseEvent.kt
```diff
@@ -0,0 +1,34 @@
...
+
+package com.ritense.outbox.domain
+import com.fasterxml.jackson.databind.node.ObjectNode
+import com.ritense.valtimo.contract.utils.SecurityUtils
+import java.time.LocalDateTime
+import java.util.*
+
+abstract class BaseEvent (
+    open val id: UUID  = UUID.randomUUID(),
+    open var source: String?,
```
ivo-ritense: This can be removed from this class. It will be set when transforming to a cloud event later.

outbox/src/main/kotlin/com/ritense/outbox/domain/BaseEvent.kt
```diff
@@ -0,0 +1,34 @@
...
+    open val specversion: String,
+    open val type: String,
+    open val data: LocalDateTime = LocalDateTime.now(),
+    open val userId: String = SecurityUtils.getCurrentUserLogin() ?: "System",
```
ivo-ritense: Change default value to null and make userId and roles nullable.

outbox/src/main/kotlin/com/ritense/outbox/OutboxService.kt
```diff
@@ -26,11 +27,14 @@ import java.util.UUID
...
     private val outboxMessageRepository: OutboxMessageRepository,
     private val objectMapper: ObjectMapper,
+    private val springApplicationName: String?,
+    private val valtimoSystemUserId: String?
 ) {
 
     @Transactional(propagation = Propagation.MANDATORY)
-    open fun send(message: Any) {
-        send(objectMapper.valueToTree(message))
+    open fun send(message: BaseEvent) {
```
ivo-ritense: Set userId and roles based on UserProvider if they are null

outbox/src/main/kotlin/com/ritense/outbox/domain/BaseEvent.kt
```diff
@@ -0,0 +1,34 @@
...
+ * https://joinup.ec.europa.eu/collection/eupl/eupl-text-eupl-12
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" basis,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.ritense.outbox.domain
```
Klaas-Ritense, line 17: The BaseEvent should go to the `contract` module so that all other modules can make use of it.
 this means the `outbox` module will need a dependency on `contract` after all.

@ivo-ritense thoughts?

ivo-ritense: I don't think so. First of all the contract module is a bit of an anti-pattern. But it doesn't make sense in this case, because the outbox module is needed in either case to be able to send the event. So moving it to the contract module will only make this module unavailable to use without the contract module, while not proving any dependency benefit in any of the other modules.

ivo-ritense (approved):

PR #994, opened 2023-10-31
Feature/multi tenancy outbox

Tom-Ritense:
Outbox for Valtimo


outbox-v2/src/main/kotlin/com/ritense/outbox/publisher/MessagePublisher.kt
```diff
@@ -0,0 +1,23 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.ritense.outbox.publisher
+
+import com.ritense.outbox.domain.OutboxMessage
+
+interface MessagePublisher {
+    fun publish(message: OutboxMessage)
```
laurens-ritense, line 22: Why no log statement here?

Tom-Ritense: It's a interface have a look at the defaultMessagePublisher

laurens-ritense: 👍

PR #1002, opened 2023-11-03
87074 Add configuration property to disable outbox

theo-ritense:
- Added `@ConditionalOnOutboxEnabled` annotation that checks the `valtimo.outbox.enabled` property and can be used by custom message publishers as well
- Moved `LoggingMessagePublisher` to its own configuration so that the priority can be set to lowest. This makes it easier/less error prone for custom implementations to override the `LoggingMessagePublisher`.


outbox/src/test/kotlin/com/ritense/outbox/config/condition/OnOutboxEnabledConditionIntTest.kt
Klaas-Ritense, line 1: Missing copyright header

theo-ritense: fixed

marijnritense (approved):

PR #1006, opened 2023-11-07
Feature/87379 add json schema document events

theo-ritense:


document/src/main/java/com/ritense/document/web/rest/impl/JsonSchemaDocumentSnapshotResource.java
```diff
@@ -39,6 +40,7 @@
 
 @RestController
 @RequestMapping(value = "/api", produces = APPLICATION_JSON_UTF8_VALUE)
+@ConditionalOnBean(DocumentSnapshotService.class)
```
marijnritense, line 43: Is this technically a breaking change? If so, let's not forget to mention it in documentation

document/src/main/kotlin/com/ritense/document/event/DocumentAssigned.kt
```diff
@@ -0,0 +1,27 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.ritense.document.event
+
+import com.fasterxml.jackson.databind.node.ObjectNode
+import com.ritense.outbox.domain.BaseEvent
+
+class DocumentAssigned(documentId: String, documentContent: ObjectNode) : BaseEvent(
```
marijnritense, line 22: Alternatively, this could accept a document as the arguments and then take the ID and content from the document.

marijnritense (approved):

document/src/main/java/com/ritense/document/web/rest/impl/JsonSchemaDocumentSnapshotResource.java
theo-ritense: No, because the bean cannot be created without the DocumentSnapshotService anyway. It would just crash.

document/src/main/kotlin/com/ritense/document/event/DocumentAssigned.kt
theo-ritense: I've thought of that, but that was not would not match the style of the events already there, and and we'd need an `ObjectMapper` in the event class. We can still refactor it later on if we change our mind.

PR #1022, opened 2023-11-20
#89007 make valtimo.keycloak.client property optional

Klaas-Ritense:


keycloak-iam/src/main/java/com/valtimo/keycloak/service/KeycloakService.java
```diff
@@ -62,6 +62,9 @@ public RolesResource clientRolesResource() {
     }
 
     public String getClientId() {
+        if (clientName.isBlank()) {
```
theo-ritense, line 65: By the looks of it, clientName can also be `null`, which will throw a NPE

keycloak-iam/src/main/java/com/valtimo/keycloak/service/KeycloakUserManagementService.java
```diff
@@ -244,13 +244,14 @@ private List<RoleRepresentation> getRolesFromUser(UserRepresentation userReprese
...
             .roles().realmLevel().listAll();
-        var clientRoles = keycloakService
-            .usersResource()
-            .get(userRepresentation.getId())
-            .roles().clientLevel(keycloakService.getClientId()).listAll();
-        var roles = new ArrayList<RoleRepresentation>();
-        roles.addAll(realmRoles);
-        roles.addAll(clientRoles);
+        var roles = new ArrayList<>(realmRoles);
+        if (!clientName.isBlank()) {
```
theo-ritense, line 248: `clientName` can be `null`

keycloak-iam/src/main/java/com/valtimo/keycloak/service/KeycloakService.java
Klaas-Ritense: No it can't by null. This is how the clientName is checked in all other places

theo-ritense: I disagree on the fact that we're relying on other classes/configuration to ensure null-safety in this one. It would be better to check for nulls in the constructor. 

However, the result would be the same: some exception would be thrown at runtime, so I'm letting it pass.

theo-ritense (approved):

PR #1043, opened 2023-11-24
fix: note deleted event

mbritense:
-


ivo-ritense (approved):

ivo-ritense:
After looking at it again this change doesn't seem in line with the rest of the application

ivo-ritense (changes requested):
No sure we want to remove the resultType

PR #1044, opened 2023-11-24
#90749 slow task user candidates

Klaas-Ritense:


contract/src/main/kotlin/com/ritense/valtimo/contract/authentication/NamedUser.kt
```diff
@@ -18,14 +18,34 @@ package com.ritense.valtimo.contract.authentication
 
 data class NamedUser(
     val id: String,
+    val email: String?,
```
theo-ritense, line 21: Please make this the last argument with a default of null, to make it non-breaking (in most cases)

Klaas-Ritense: Added a constructor

theo-ritense (approved):

PR #1052, opened 2023-11-27
Added more exports

theo-ritense:
88238 - Export process document associations
88070 - Export process definitions
88071 - Export decision definitions
88072 - Export process links

### DIFF
--- a/core/src/test/kotlin/com/ritense/valtimo/export/ProcessDefinitionExporterIntTest.kt
+++ b/core/src/test/kotlin/com/ritense/valtimo/export/ProcessDefinitionExporterIntTest.kt
@@ -1,3 +1,19 @@
+/*
+ * Copyright 2015-2023 Ritense BV, the Netherlands.
+ *
+ * Licensed under EUPL, Version 1.2 (the "License");
+ * you may not use this file except in compliance with the License.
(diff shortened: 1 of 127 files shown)


core/src/test/kotlin/com/ritense/valtimo/export/ProcessDefinitionExporterIntTest.kt
```diff
@@ -0,0 +1,81 @@
...
+
+@Transactional(readOnly = true)
+class ProcessDefinitionExporterIntTest @Autowired constructor(
+    private val repositoryService: RepositoryService,
+    private val camundaRepositoryService:CamundaRepositoryService,
+    private val processDefinitionExporter: ProcessDefinitionExporter
+) : BaseIntegrationTest() {
+
+    @Test
+    fun `should export process definition`(): Unit = runWithoutAuthorization {
```
marijnritense: Can you also make a test that covers a non-existing DMN definition? And what if there's no business rule in the process at all?

theo-ritense: done

marijnritense (approved):

PR #1056, opened 2023-11-28
Feature/multi tenancy outbox v2

nielserkamp-ritense:


outbox-v2/src/main/kotlin/com/ritense/outbox/service/DomainEventOutboxService.kt
```diff
@@ -1,47 +1,34 @@
...
- * https://joinup.ec.europa.eu/collection/eupl/eupl-text-eupl-12
- *
- * Unless required by applicable law or agreed to in writing, software
- * distributed under the License is distributed on an "AS IS" basis,
- * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
- * See the License for the specific language governing permissions and
- * limitations under the License.
- */
-
 package com.ritense.outbox.service
```
Tom-Ritense, line 17: pls ring back the license otherwise the build will fail. All classes need this event tests.

nielserkamp-ritense: Adjusted

outbox-v2/src/main/kotlin/com/ritense/outbox/publisher/MessagePublishingFailed.kt
```diff
@@ -14,9 +14,6 @@
...
 
-package com.ritense.outbox.domain
+package com.ritense.outbox.publisher
 
-import java.io.Serializable
-
-interface DomainEvent: Serializable {
-}
\ No newline at end of file
+open class MessagePublishingFailed(message: String) : RuntimeException(message)
```
sirolf2009: Doesn't seem to be used

sirolf2009: and also, adding a (optional) cause would be appreciated

nielserkamp-ritense: An exception class can always be added as here you have a class you can use, at it is a library / module in the end (not an end application).

I copied it from Valtimo so didn't adjust it, now did though ;) because you are right should have more options.

sirolf2009 (approved):

PR #1066, opened 2023-11-30
Changed pbac relation to where in subquery type

ivo-ritense:


ivo-ritense:
Cherry-picked changes to 11.0 instead.

PR #1073, opened 2023-12-01
#85463 zaak value resolver

Klaas-Ritense:
Screenshot 2023-12-01 at 16 30 49


zgw/zaken-api/src/main/kotlin/com/ritense/zakenapi/resolver/ZaakStatusValueResolverFactory.kt
```diff
@@ -0,0 +1,71 @@
...
+
+    override fun handleValues(
+        processInstanceId: String,
+        variableScope: VariableScope?,
+        values: Map<String, Any>
+    ) {
+        TODO()
+    }
+
+    override fun getResolvedValue(documentId: UUID, field: String): Any? {
```
ivo-ritense: The only issue i have with this is that this method might be called multiple times (if we need multiple zaakstatus variables) and then the zaakstatus gets retrieved multiple times as well. We try to bundle the fields based on resolver type so that we can minimize this sort of thing.

Klaas-Ritense: Changed

ivo-ritense (approved):

PR #1077, opened 2023-12-05
87961 Create import endpoint

theo-ritense:


case/src/main/kotlin/com/ritense/case/security/config/CaseHttpSecurityConfigurer.kt
```diff
@@ -51,6 +51,7 @@ class CaseHttpSecurityConfigurer : HttpSecurityConfigurer {
                 .antMatchers(DELETE, "/api/management/v1/case-definition/{caseDefinitionName}/tab/{tabKey}").hasAuthority(ADMIN)
                 .antMatchers(GET, "/api/management/v1/case-definition/{caseDefinitionName}/tab").hasAuthority(ADMIN)
                 .antMatchers(GET, "/api/management/v1/case/{caseDefinitionName}/{caseDefinitionVersion}/export").hasAuthority(ADMIN)
+                .antMatchers(GET, "/api/management/v1/case/import").hasAuthority(ADMIN)
```
marijnritense:
```suggestion
                .antMatchers(POST, "/api/management/v1/case/import").hasAuthority(ADMIN)
```

import/src/main/kotlin/com/ritense/import/ValtimoImportService.kt
```diff
@@ -0,0 +1,172 @@
...
+
+    init {
+        //Check for duplicate import types in the list of importers
+        val wImporters = importers.map { WrappedImporter(it) }.toSet()
+        if (wImporters.size != importers.size) {
+            val duplicatedTypes = wImporters.filter { wImporter ->
+                importers.count { wImporter.type() == it.type() } > 1
+            }.map { it.type() }
+            throw ImportServiceException(
+                "Multiple importers of the same type provided: [${duplicatedTypes.joinToString()}]"
```
marijnritense, line 39: Small "concern": By leaving the description of the exception completely flexible, you're diluting the meaning of the exception somewhat. I think in general we should try to make more specific exceptions, with only some parts being variable. Thoughts?

theo-ritense: The main goal for the typed exception was to be able to throw a status 400 instead of a 500. Because this will be thrown at startup, that scenario doesn't even apply.

But while in general you are right, I don't like having I don't know how many specific Exception classes that we handle the same regardless.

import/src/test/kotlin/com/ritense/import/ValtimoImportServiceTest.kt
```diff
@@ -0,0 +1,111 @@
...
+                )
+            )
+        }
+
+        assertThat(exception.message)
+            .isEqualTo("Importer dependencies could not be resolved or contain a cyclic reference! Error occurred after: [1]")
+    }
+
+    @Test
+    fun `should import files from zip in the right order`() {
```
marijnritense, line 67: Can you add a test scenario, where the dependencies could not be fulfilled in the import?

theo-ritense: I removed the check because that state can no longer be reached.

marijnritense (approved):

PR #1079, opened 2023-12-05
Fix deadlock

Klaas-Ritense:


theo-ritense (commented):
While this works, this strictly speaking removes a feature. This fix opens up the possibility that 2 parallel job/threads can change the document, which might lead to data loss.

The right solution would be to use transaction locking the right way (`FOR UPDATE SKIP LOCKED`?)

document/src/main/java/com/ritense/document/domain/impl/JsonSchemaDocument.java
```diff
@@ -197,7 +197,7 @@ public static CreateDocumentResultImpl create(
      * @param versionCheck    The version on which the content was based on (in other words: this's ver)
      * @return Object representing the result of the operation (either resulting document or errors)
      */
-    public synchronized ModifyDocumentResultImpl applyModifiedContent(
+    public ModifyDocumentResultImpl applyModifiedContent(
```
Klaas-Ritense, line 200: @theo-ritense 
> This fix opens up the possibility that 2 parallel job/threads can change the document, which might lead to data loss.

No it doesn't. It will then throw an error in one of the 2 parallel jobs: [ConflictedDocumentVersion](https://github.com/valtimo-platform/valtimo-backend-libraries/blob/development/11.2.0/document/src/main/java/com/ritense/document/domain/impl/JsonSchemaDocument.java#L210)

theo-ritense: This actually proves my point. I can still call the `applyModifiedContent` in parallel on the same version. This check only works properly after the transaction finishes, which, with some incorrectly modeled processes, can take quite long. 

Before the transaction finishes, one can still retrieve the same version and perform an update on it.

I know: it's hypothetical, but certainly not impossible.

Tom-Ritense: @theo-ritense @anna-ritense I spoke to Anna yesterday about this topic. And in the implementations we build in our own project this is not hypothetical but a reality. I would advice that we create a new PR that addresses this issue because its a nasty bug when it occurs. Data integrity bugs are a pain to find out. @theo-ritense can we discuss a solution? Main contender to fix this is using a version annotation provided by the JPA framework. This is a optimistic locking strategy. https://www.baeldung.com/jpa-optimistic-locking

ivo-ritense (approved):

PR #1095, opened 2023-12-10
Use last form flow step instance when multiple instances exists

erik-ritense:
See TP 93043


marijnritense (changes requested):
I'm not sure if this actually solves the issue. It seems like you're working around it. Blocking this PR from being merged until I've discussed this with the rest of the team.

marijnritense:
Closing PR, should be fixed with [this pr](https://github.com/valtimo-platform/valtimo-backend-libraries/pull/1116)

PR #1118, opened 2023-12-18
#85463 Improved error messages

Klaas-Ritense:


zgw/zaken-api/src/main/kotlin/com/ritense/zakenapi/service/ZaakDocumentService.kt
```diff
@@ -82,4 +87,14 @@ class ZaakDocumentService(
         return plugin?.getZaak(url)
     }
 
+    fun getZaakByDocumentIdOrThrow(documentId: UUID): ZaakResponse {
+        val url = zaakUrlProvider.getZaakUrl(documentId)
+        val plugin = pluginService.createInstance(
+            ZakenApiPlugin::class.java,
+            ZakenApiPlugin.findConfigurationByUrl(url)
+        )
+            ?: throw IllegalStateException("Missing plugin configuration of type '${ZakenApiPlugin.PLUGIN_KEY}' for url '$url'")
```
theo-ritense, line 96: If you want it to be correct, you have to call `PluginService.findPluginConfiguration(clazz: Class<T>, configurationFilter: (JsonNode) -> Boolean)` first, because the error could be caused by something different.

zgw/zaken-api/src/main/kotlin/com/ritense/zakenapi/resolver/ZaakStatusValueResolverFactory.kt
```diff
@@ -59,14 +59,15 @@ class ZaakStatusValueResolverFactory(
...
+            ?: throw IllegalStateException("Missing plugin configuration of type '${ZakenApiPlugin.PLUGIN_KEY}' for url '$url'")
     }
 
     private fun getCatalogiApiPlugin(statusTypeUrl: URI): CatalogiApiPlugin {
         return pluginService.createInstance(
             CatalogiApiPlugin::class.java,
             CatalogiApiPlugin.findConfigurationByUrl(statusTypeUrl)
         )
-            ?: throw IllegalStateException("Missing plugin of type '${CatalogiApiPlugin.PLUGIN_KEY}' for statusType '$statusTypeUrl'")
+            ?: throw IllegalStateException("Missing plugin configuration of type '${CatalogiApiPlugin.PLUGIN_KEY}' for statusType '$statusTypeUrl'")
```
theo-ritense, line 71: If you want it to be correct, you have to call `PluginService.findPluginConfiguration(clazz: Class<T>, configurationFilter: (JsonNode) -> Boolean)` first, because the error could be caused by something different.

Klaas-Ritense: I've looked into it. The error here is already correct. The `PluginService.findPluginConfiguration` is already being called by the `PluginService.createInstance` which will return null when no plugin configuration is found. Thats what the error here catches

theo-ritense: Yes, but it will also throw an error when the configuration is found, but an instance cannot be created. For example, an error in the factory, or configuration reflection.

Klaas-Ritense: Yes it can throw an error there. Why is that a problem?

theo-ritense: Because the message does not reflect the actual error in that case.

Klaas-Ritense: But I haven't changed the original error message

ivo-ritense (approved):

PR #1133, opened 2023-12-29
#93754 Support Keycloak group roles

Klaas-Ritense:


theo-ritense (commented):
~~I don't think we've covered every use-case, since `getRolesFromUser(UserRepresentation userRepresentation)` was not changed.~~

~~We might also be able to improve performance by using the `briefRepresentation` param on some of these calls, eliminating the need to do calls for nested data. But I'm not sure, since calling the API with `briefRepresentation` might be slow on its own.~~

fixed

keycloak-iam/src/main/java/com/valtimo/keycloak/service/KeycloakUserManagementService.java
```diff
@@ -266,13 +276,13 @@ private List<RoleRepresentation> getRolesFromUser(UserRepresentation userReprese
...
                 .get(userRepresentation.getId())
-                .roles().realmLevel().listAll();
+                .roles().realmLevel().listEffective(true);
             var roles = new ArrayList<>(realmRoles);
             if (!clientName.isBlank()) {
                 var clientRoles = keycloakService
                     .usersResource(keycloak)
                     .get(userRepresentation.getId())
-                    .roles().clientLevel(keycloakService.getClientId(keycloak)).listAll();
+                    .roles().clientLevel(keycloakService.getClientId(keycloak)).listEffective(true);
```
Klaas-Ritense, line 286: @theo-ritense I've made some changes.

Here the `listEffective(true)` will make sure to also retrieve group roles. I've also enabled `briefRepresentation` where possible.

marijnritense (approved):

ivo-ritense (approved):

PR #1135, opened 2024-01-02
#87470 Show Process Version on Progress Tab

Klaas-Ritense:


process-document/src/main/java/com/ritense/processdocument/web/rest/ProcessDocumentResource.java
```diff
@@ -128,11 +128,11 @@ public ResponseEntity<ProcessDocumentDefinition> getProcessDocumentDefinition(
     }
 
     @GetMapping("/v1/process-document/instance/document/{documentId}")
-    public ResponseEntity<List<? extends ProcessDocumentInstance>> findProcessDocumentInstances(
+    public ResponseEntity<List<ProcessDocumentInstanceDto>> findProcessDocumentInstances(
```
ivo-ritense: Isn't this a breaking change? Maybe we should create a v2

Klaas-Ritense: Changed it. But v2 isn't needed

ivo-ritense (approved):

PR #1147, opened 2024-01-08
Lost update fix via Version annotation for documents

marijnritense:
Added version column to document with a default value. This is a breaking change as the old version field has been replaced.


ivo-ritense (approved):

document/src/main/java/com/ritense/document/domain/impl/event/JsonSchemaDocumentCreatedEvent.java
```diff
@@ -44,16 +44,14 @@ public JsonSchemaDocumentCreatedEvent(
...
+        JsonSchemaDocumentDefinitionId definitionId
     ) {
         super(id, origin, occurredOn, user);
         assertArgumentNotNull(documentId, "documentId is required");
         assertArgumentNotNull(definitionId, "definitionId is required");
-        assertArgumentNotNull(version, "version is required");
         this.documentId = documentId;
         this.definitionId = definitionId;
-        this.version = version;
+        this.version = null;
```
Klaas-Ritense, line 54: Can't this version be removed if its no longer used?

PR #1153, opened 2024-01-11
#94616 Bug in plugin action property

Klaas-Ritense:
and other fixes


plugin-valtimo/src/test/kotlin/com/ritense/valtimo/TestPlugin.kt
```diff
@@ -33,7 +33,11 @@ class TestPlugin {
         description = "This is a test action",
         activityTypes = [SERVICE_TASK_START]
     )
-    fun testAction(@PluginActionProperty testActionProperty: String) {
-        //do nothing
+    fun testAction(@PluginActionProperty attachmentIds: List<String>?) {
+        Companion.attachmentIds = attachmentIds
+    }
+
+    companion object {
```
theo-ritense, line 40: The `companion object` behaves a bit like a static. So this test might behave differently when multiple tests use this.

Better to make it an instance var, so that it resets when a new instance is created.

Klaas-Ritense: Making an instance var is not possible in this case. Because every time this TestPlugin is retrieved, you get a new instance

ivo-ritense (dismissed):

theo-ritense (approved):

plugin-valtimo/src/test/kotlin/com/ritense/valtimo/TestPlugin.kt
theo-ritense: You could replace it with some callback functionality which you can inject from your test. But we'll fix that when we get there.

PR #1162, opened 2024-01-18
#68171 Migrate camunda tasks to use User ID instead of email address

Klaas-Ritense:


core/src/main/java/com/ritense/valtimo/service/CamundaTaskService.java
```diff
@@ -161,13 +163,22 @@ public CamundaTask findTaskById(String taskId) {
...
+        var assignee = userManagementService.findNamedUserByEmail(assigneeEmail)
+            .orElseThrow(() -> new IllegalStateException("Error. No registered user found with email: " + assigneeEmail));
+        assign(taskId, assignee.getId());
+    }
+
     @Transactional
     public void assign(String taskId, String assignee) throws IllegalStateException {
         if (assignee == null) {
             unassign(taskId);
+        } else if (assignee.matches("[a-zA-Z0-9._-]+@[a-zA-Z0-9.-]+\\.[a-zA-Z]{2,4}")) {
```
theo-ritense: Matching email addresses with a regex is [pretty tricky](https://emailregex.com/index.html). For instance, your version does not support a `+` before the `@`.

keycloak-iam/src/main/java/com/valtimo/keycloak/liquibase/changelog/ChangeLog20240116MigrateTaskAssigneeEmailToUserId.kt
```diff
@@ -0,0 +1,192 @@
...
+
+        val connection = database.connection as JdbcConnection
+        val statement = connection.prepareStatement("SELECT id_,assignee_ FROM act_ru_task")
+        val result = statement.executeQuery()
+
+        while (result.next()) {
+            val taskId = result.getString("id_")
+            val taskAssigneeEmail = result.getString("assignee_")
+            if (taskAssigneeEmail != null) {
+                if (!taskAssigneeEmail.matches("[a-zA-Z0-9._-]+@[a-zA-Z0-9.-]+\\.[a-zA-Z]{2,4}".toRegex())) {
```
theo-ritense: See other comment about email matching

keycloak-iam/src/main/java/com/valtimo/keycloak/liquibase/changelog/ChangeLog20240116MigrateTaskAssigneeEmailToUserId.kt
```diff
@@ -0,0 +1,192 @@
...
+        listOf(name, toCamelCase(name), toSnakeCase(name), toJoinedLowercase(name)).forEach {
+            val ymlValue = getApplicationYmlProperty(it)
+            if (ymlValue != null) {
+                return ymlValue
+            }
+        }
+        throw IllegalStateException("Failed to find application property: '$name'")
+    }
+
+    private fun getApplicationYmlProperty(propertyPath: String): String? {
```
theo-ritense: What if an implementation uses `application.properties`? Can't we somehow use the property resolvers of Spring?

keycloak-iam/src/main/java/com/valtimo/keycloak/liquibase/changelog/ChangeLog20240116MigrateTaskAssigneeEmailToUserId.kt
Klaas-Ritense: It now uses the email validator from keycloak. This should be perfect, since this changeset is made to migrate keycloak emails.

core/src/main/java/com/ritense/valtimo/service/CamundaTaskService.java
Klaas-Ritense: I can't use the keycloak validation here. But this extra check is only there for developers that forgot to change the assign function to now use a keycloak UUID instead of an email.

keycloak-iam/src/main/java/com/valtimo/keycloak/liquibase/changelog/ChangeLog20240116MigrateTaskAssigneeEmailToUserId.kt
Klaas-Ritense: `application.properties` wont work right now. The problem is, Liquibase changesets are run before most Spring beans are available. But Ill look onto it

core/src/main/java/com/ritense/valtimo/service/CamundaTaskService.java
theo-ritense: You could instead use the `EmailValidator` from Apache commons, or the regex referenced by [OWASP](https://owasp.org/www-community/OWASP_Validation_Regex_Repository):
`^[a-zA-Z0-9_+&*-]+(?:\\.[a-zA-Z0-9_+&*-]+)*@(?:[a-zA-Z0-9-]+\\.)+[a-zA-Z]{2,}$`

marijnritense (approved):

ivo-ritense (approved):

PR #1164, opened 2024-01-18
Increased message size

Tom-Ritense:


outbox-v2/src/main/kotlin/com/ritense/outbox/domain/OutboxMessage.kt
```diff
@@ -31,7 +31,7 @@ class OutboxMessage(
     @Column(name = "id", columnDefinition = "BINARY(16)", updatable = false)
     val id: UUID,
 
-    @Column(name = "message", columnDefinition = "TEXT", updatable = false)
+    @Column(name = "message", updatable = false)
```
laurens-ritense, line 34: Why no columnDefinition?

PR #1165, opened 2024-01-18
Different gradle tasks for integration testing depending on the DBMS

marijnritense:


gradle/test.gradle
```diff
@@ -26,8 +26,9 @@ test {
     }
 }
 
-tasks.register("integrationTesting", Test) {
+tasks.register("integrationTestingPostgresql", Test) {
```
theo-ritense, line 33: Please add an alias for backwards compatibility with the pipeline and ease of use.
```suggestion
tasks.register("integrationTesting") {
    dependsOn("integrationTestingPostgresql")
}

tasks.register("integrationTestingPostgresql", Test) {
```

docker-resources/docker-compose-base-test-mysql.yml
```diff
@@ -0,0 +1,66 @@
...
+            --sql-mode=NO_ENGINE_SUBSTITUTION
+        #ports:
+           # - "5432:5432" #needs overriding "XXXX:5432" XXXX=your-own-port-per-module-test
+        environment:
+            MYSQL_ROOT_PASSWORD: password
+            MYSQL_DATABASE: valtimo #needs overriding
+            MYSQL_USER: valtimo
+            MYSQL_PASSWORD: password
+# Ports taken
+#   url: jdbc:postgresql://localhost:3301/form-test
```
theo-ritense: I would not expect jdbc:postgresql urls here. Maybe move the 'Ports taken' somewhere else, or only keep it at the postgres declaration.

objects-api/src/test/resources/config/document/application-mysql.yml
theo-ritense, line 1: why do we need these configs in `/config/document/` ?

outbox/outbox-rabbitmq/src/test/kotlin/com/ritense/outbox/rabbitmq/RabbitMessagePublisherIntTest.kt
```diff
@@ -54,7 +55,8 @@ class RabbitMessagePublisherIntTest {
     @Nested
     @ActiveProfiles("invalidrouting")
     inner class InvalidRouting @Autowired constructor(
-        val springCloudMessagePublisher: RabbitMessagePublisher
+        val springCloudMessagePublisher: RabbitMessagePublisher,
+        val environment: Environment
```
theo-ritense: I this a remnant of your attempt to fix some tests?

marijnritense: It is, thanks for pointing it out. I removed it.

objects-api/src/test/resources/config/document/application-mysql.yml
marijnritense: We don't, I removed them

docker-resources/docker-compose-base-test-mysql.yml
marijnritense: I changed `postgresql` to `mysql` for now.

theo-ritense (approved):

PR #1170, opened 2024-01-23
Feature/sonarqube code smells

theo-ritense:


contract/src/main/java/com/ritense/valtimo/contract/security/jwt/TokenAuthenticator.java
```diff
@@ -19,10 +19,10 @@
 import io.jsonwebtoken.Claims;
 import org.springframework.security.core.Authentication;
 
-public abstract class TokenAuthenticator {
+public interface TokenAuthenticator {
```
marijnritense, line 22: Don't forget to also document this as a breaking change

core/src/main/java/com/ritense/valtimo/camunda/repository/CustomRepositoryServiceImpl.java
```diff
@@ -35,11 +35,6 @@ public CustomRepositoryServiceImpl(final ApplicationEventPublisher applicationEv
         this.applicationEventPublisher = applicationEventPublisher;
     }
 
-    @Override
-    public DeploymentWithDefinitions deployWithResult(DeploymentBuilderImpl deploymentBuilder) {
```
marijnritense, line 39: Why was this removed?

theo-ritense: Because it overrides a method and then only calls the `super`.

theo-ritense: Because it overrides a method, but only calls `super`. So nothing changed effectively.

contract/src/main/java/com/ritense/valtimo/contract/security/jwt/TokenAuthenticator.java
theo-ritense: I made a not of it in the story.

marijnritense (approved):

PR #1173, opened 2024-01-24
Changed actuator health status for unhealthy - but not down - services and applications

Hildebrand-Ritense:
Changed status of CamundaIncident health indicator to 'restricted' when unhealthy.

Same for ValtimoHealthAggregator, changing from 'unknown' to 'restricted' to better reflect the application's state and align with how Spring Boot Admin processes the statuses down the line.

See https://github.com/codecentric/spring-boot-admin/issues/3095

Visually this results in the following:
Going from this, where the indicator is reporting 'unknown':
 
sba-status-unknown

to this, where the indicator is reporting 'restricted':

sba-status-restricted

Also, added security configuration exposing /health/liveness and /health/readiness so the k8s health probes are accessible.


gradle.properties
```diff
@@ -97,4 +97,4 @@ snakeYamlVersionOverride=2.0
 
 # project version
 
-projectVersion=11.2.0
+projectVersion=11.2.1.1-SNAPSHOT
```
ThomasMinkeRitense: Can you please revert this change?

Hildebrand-Ritense: yes, done

Klaas-Ritense (dismissed):

ThomasMinkeRitense (approved):

PR #1175, opened 2024-01-25
Fix sonarqube pipeline task

theo-ritense:


build.gradle
```diff
@@ -82,7 +82,7 @@ configurations {
 }
 
 allprojects {
-    if (!(project.name.endsWith("-dependencies") || project.name.endsWith("-dependency-versions"))) {
```
ThomasMinkeRitense, line 85: Why don't we exclude the -dependencies modules any more?

theo-ritense: Because by excluding them, we disable the dependency management of the Spring Boot plugin, hence SonarQube can't determine the Spring dependency versions. The other workaround was to add the `valtimo-dependency-versions` as a dependency to `valtimo-(gzac-)dependencies`, but that would turn them into a BOM/POM yet again.

PS: This is not the same block as `subprojects`. `-dependencies` are still handled differently there

ThomasMinkeRitense (approved):

PR #1176, opened 2024-01-26
#97911 Fix failing MySQL integration tests

Klaas-Ritense:


audit/src/test/resources/config/application-mysql.yml
```diff
@@ -6,8 +6,10 @@ spring:
         password: password
         hikari:
             auto-commit: false
+            data-source-properties:
+                serverTimezone: UTC
```
Klaas-Ritense, line 10: Fix for timezone issues on mysql

audit/src/test/resources/config/application-mysql.yml
```diff
@@ -6,8 +6,10 @@ spring:
...
     jpa:
-        database-platform: org.hibernate.dialect.MySQLDialect
+        database-platform: org.hibernate.dialect.MySQL8Dialect
```
Klaas-Ritense, line 12: Mysql query '...for update skip locked' for outboxmessages is only supported on mysql 8+

core/src/test/kotlin/com/ritense/valtimo/camunda/repository/CamundaHistoricProcessInstanceSpecificationHelperIntTest.kt
```diff
@@ -130,9 +145,11 @@ class CamundaHistoricProcessInstanceSpecificationHelperIntTest @Autowired constr
     @Transactional
     @WithMockUser(username = TEST_USER, authorities = [ADMIN])
     fun byEndTimeAfter() {
+        Thread.sleep(1000)
```
Klaas-Ritense, line 148: Mysql saves datatype 'datetime' without miliseconds. Unlike postgres. Comparisons involving datetime on MySQL like `time1 < time2` will often fail because the time1 and time2 are equal

outbox/src/main/kotlin/com/ritense/outbox/OutboxMessageRepository.kt
```diff
@@ -31,7 +31,7 @@ interface OutboxMessageRepository : JpaRepository<OutboxMessage, UUID> {
      */
     @QueryHints(value = [QueryHint(name = JPA_LOCK_TIMEOUT, value = SKIP_LOCKED)])
     @Lock(PESSIMISTIC_WRITE)
-    fun findTopByOrderByCreatedOnAsc(): OutboxMessage?
+    fun findFirstBy(): OutboxMessage?
```
Klaas-Ritense: On Mysql this query: 
`SELECT * from outbox_message ORDER BY created_on ASC LIMIT 1 FOR UPDATE SKIP LOCKED`
Will lock almost all rows in the table. Not sure why. I think its due to wrongly 'gap-locking' some rows.

On Postgres 'gap-locking' also happens but is done correctly

core/src/main/java/com/ritense/valtimo/service/CamundaTaskService.java
```diff
@@ -269,6 +269,7 @@ public void complete(String taskId) {
         Hibernate.initialize(task.getVariableInstances());
         Hibernate.initialize(task.getIdentityLinks());
         entityManager.detach(task);
+        task.getIdentityLinks().forEach(entityManager::detach);
```
Klaas-Ritense: Valtimo sometimes tries to save the identitylinks which can result in errors

theo-ritense: Can you add that as a comment to the code?

core/src/test/kotlin/com/ritense/valtimo/camunda/repository/CamundaHistoricProcessInstanceSpecificationHelperIntTest.kt
theo-ritense: I get the problem, but I hate the solution. I'll take a closer look to see if I can brew something without having to block the thread.

outbox/src/main/kotlin/com/ritense/outbox/OutboxMessageRepository.kt
theo-ritense: What does `findFirstBy` create as a query? By what field does it find the first?

Klaas-Ritense: The generated query will be:
``` SELECT * from outbox_message LIMIT 1 FOR UPDATE SKIP LOCKED```
There is no ordering anymore because that locked the entire MySql table. But the messages are no longer consumed in the correct order. Which is a big loss imo

core/src/test/kotlin/com/ritense/valtimo/camunda/repository/CamundaHistoricProcessInstanceSpecificationHelperIntTest.kt
Klaas-Ritense: I looked into https://github.com/awaitility/awaitility. but it doesn't work neatly in this case

core/src/main/java/com/ritense/valtimo/service/CamundaTaskService.java
Klaas-Ritense: Added

outbox/src/main/kotlin/com/ritense/outbox/OutboxMessageRepository.kt
theo-ritense: I'd like to see a separate repository for Postgres and Mysql to solve this. I can live with MySQL not having this feature, but Postgres should still have it.

core/src/test/kotlin/com/ritense/valtimo/camunda/repository/CamundaHistoricProcessInstanceSpecificationHelperIntTest.kt
theo-ritense: I tried awaitily as well and concluded the same. But I didn't have enough time to properly check an alternative yet.

theo-ritense: I tried a few things, but it's finicky to make a solution work with both MySQL and PostgreSQL. One of the issues is that we are not testing a real `byEndTimeAfter', but a `byEndTimeEqualOrAfter`.

We could think of changing the mysql column types: https://stackoverflow.com/a/26299379, but I that's stretching the goal of this PR a bit much. I'll add a new story.

theo-ritense (approved):

PR #1189, opened 2024-01-31
new endpoint to retrieve template

marijnritense:


document/src/main/java/com/ritense/document/web/rest/impl/JsonSchemaDocumentDefinitionResource.java
```diff
@@ -61,6 +61,22 @@ public ResponseEntity<Page<? extends DocumentDefinition>> getDocumentDefinitions
         return ok(documentDefinitionService.findAll(fixPageable(pageable)));
     }
 
+    @Override
+    public ResponseEntity<String> getDocumentDefinitionTemplate() {
```
theo-ritense: I guess the story was not clear. The `$id` and `title` should be provided by the FE and replaced in the template. I believe that also means this should be a POST?

marijnritense: Fixed

document/src/main/java/com/ritense/document/domain/impl/template/DocumentDefinitionTemplateRequestDto.java
```diff
@@ -0,0 +1,20 @@
+/*
+ * Copyright 2015-2023 Ritense BV, the Netherlands.
```
theo-ritense:
```suggestion
 * Copyright 2015-2024 Ritense BV, the Netherlands.
```

document/src/main/java/com/ritense/document/web/rest/impl/JsonSchemaDocumentDefinitionResource.java
```diff
@@ -61,6 +62,22 @@ public ResponseEntity<Page<? extends DocumentDefinition>> getDocumentDefinitions
         return ok(documentDefinitionService.findAll(fixPageable(pageable)));
     }
 
+    @Override
+    public ResponseEntity<String> getDocumentDefinitionTemplate(DocumentDefinitionTemplateRequestDto requestDto) {
+        return ok(
+            """
+                {
+                    "$id": "%s.schema",
```
theo-ritense, line 70: Shouldn't we sanatize the input? I could break this JSON by sending `xyz"`. Best would be to use Jackson to set/add a TextNode I think.

document/src/main/java/com/ritense/document/domain/impl/template/DocumentDefinitionTemplateRequestDto.java
marijnritense: Updated

document/src/main/java/com/ritense/document/web/rest/impl/JsonSchemaDocumentDefinitionResource.java
marijnritense: Added sanitization

theo-ritense (approved):

PR #1196, opened 2024-02-05
#99185 Deprecate connector framework

Klaas-Ritense:


plugin/src/main/kotlin/com/ritense/plugin/service/PluginService.kt
```diff
@@ -536,6 +536,14 @@ class PluginService(
         return pluginConfiguration?.let { createInstance(it) as T }
     }
 
+    fun <T> createInstances(clazz: Class<T>): List<T> {
```
theo-ritense: Why was this added if it's not used? And if it's needed: can we keep the filter argument in?

Klaas-Ritense: Changed

theo-ritense (approved):

PR #1201, opened 2024-02-07
Feature/99183 internal case status

marijnritense:


document/src/main/kotlin/com/ritense/document/domain/InternalCaseStatusId.kt
```diff
@@ -0,0 +1,35 @@
...
+package com.ritense.document.domain
+
+import com.ritense.valtimo.contract.domain.AbstractId
+import jakarta.persistence.Column
+import jakarta.persistence.Embeddable
+
+
+@Embeddable
+data class InternalCaseStatusId(
+    @Column(name = "case_definition_name", length = 50, columnDefinition = "VARCHAR(50)")
```
Klaas-Ritense: The `case_definition_name` can be 256 long

document/src/main/kotlin/com/ritense/document/service/InternalCaseStatusService.kt
```diff
@@ -0,0 +1,147 @@
...
+import kotlin.jvm.optionals.getOrNull
+
+@Transactional
+class InternalCaseStatusService(
+    private val internalCaseStatusRepository: InternalCaseStatusRepository,
+    private val documentDefinitionService: DocumentDefinitionService,
+    private val authorizationService: AuthorizationService,
+) {
+    fun getInternalCaseStatuses(documentDefinitionName: String): List<InternalCaseStatus> {
+        denyManagementOperation()
```
Klaas-Ritense, line 40: Is a DENY here the right way to go, when getting the status?

marijnritense: Yes, this is an admin-only method. Another story revolves around the case-list, so it's possible this needs to be changed then.

document/src/main/kotlin/com/ritense/document/domain/InternalCaseStatusId.kt
marijnritense: fixed

ivo-ritense: As far as i can find it is only 50

Klaas-Ritense: I was wrong. It should be 50 after all

marijnritense: fixed

Klaas-Ritense (approved):

PR #1211, opened 2024-02-13
#101009 Set the initial value of uiterlijkeEinddatumAfdoening when creating a zaak

Klaas-Ritense:


zgw/catalogi-api/src/main/kotlin/com/ritense/catalogiapi/web/rest/result/ZaaktypeDto.kt
```diff
@@ -0,0 +1,34 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.ritense.catalogiapi.web.rest.result
+
+import com.ritense.catalogiapi.domain.Zaaktype
+import java.net.URI
+
+class ZaaktypeDto(
```
marijnritense, line 22: We sometimes use *ResponseDto as a pattern as well. Not saying this has to be changed, but what do you think? Should we stick to that naming? Then this class would become `ZaaktypeResponseDto`

marijnritense (approved):

zgw/catalogi-api/src/main/kotlin/com/ritense/catalogiapi/web/rest/result/ZaaktypeDto.kt
Klaas-Ritense: Might be nice to decide on a single approach. Right now we have:
- 13x *ResponseDto
- 14x *RequestDto
- 70x *Dto
- 18x *Response
- 72x *Request

PR #1217, opened 2024-02-15
feat: configure tag colors

mbritense:
closes https://ritense.tpondemand.com/entity/100629-fe-be-configure-tag-colors


document/src/main/resources/config/liquibase/changelog/20240214-add-column-color-to-internal-case-status-table.xml
Klaas-Ritense, line 1: Missing License header

Klaas-Ritense (approved):

Issue #1218, opened 2024-02-16
Bug: Documenten API client Response uses LocalDateTime, cannot handle dates with timezone info


haiko-ritense:
Error for the field 'beginRegistratie' in com.ritense.documentenapi.client.CreateDocumentResult upon deserialising:
```
Caused by: org.springframework.core.codec.DecodingException: JSON decoding error: Cannot deserialize value of type `java.time.LocalDateTime` from String "2024-02-15T14:39:54.746+01:00": Failed to deserialize java.time.LocalDateTime: (java.time.format.DateTimeParseException) Text '2024-02-15T14:39:54.746+01:00' could not be parsed, unparsed text found at index 23; nested exception is com.fasterxml.jackson.databind.exc.InvalidFormatException: Cannot deserialize value of type `java.time.LocalDateTime` from String "2024-02-15T14:39:54.746+01:00": Failed to deserialize java.time.LocalDateTime: (java.time.format.DateTimeParseException) Text '2024-02-15T14:39:54.746+01:00' could not be parsed, unparsed text found at index 23
at [Source: (io.netty.buffer.ByteBufInputStream); line: 7, column: 25] (through reference chain: com.ritense.documentenapi.client.CreateDocumentResult["beginRegistratie"])
  at org.springframework.http.codec.json.AbstractJackson2Decoder.processException(AbstractJackson2Decoder.java:242)
```

PR #1220, opened 2024-02-16
Fix issue #1218 with LocalDateTime parsing

theo-ritense:
The standard deserialiser provided by JavaTimeModule does not parse datetimes with a zone provided to a LocalDateTime. This PR overrides part of that deserialiser to fix this.

Solves #1218 .

### DIFF
--- a/core/src/main/java/com/ritense/valtimo/config/JacksonConfiguration.java
+++ b/core/src/main/java/com/ritense/valtimo/config/JacksonConfiguration.java
@@ -22,6 +22,8 @@ import com.fasterxml.jackson.datatype.jsr310.JavaTimeModule;
 import com.fasterxml.jackson.datatype.jsr310.ser.LocalDateTimeSerializer;
 import com.fasterxml.jackson.module.blackbird.BlackbirdModule;
 import com.ritense.valtimo.contract.json.serializer.PageSerializer;
+import com.ritense.valtimo.jackson.CustomLocalDateTimeDeserializer;
+import java.time.LocalDateTime;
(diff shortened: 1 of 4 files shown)


core/src/main/java/com/ritense/valtimo/config/JacksonConfiguration.java
```diff
@@ -50,7 +52,9 @@ public Jackson2ObjectMapperBuilderCustomizer jackson2ObjectMapperBuilderCustomiz
      */
     @Bean
     public JavaTimeModule javaTimeModule() {
-        return new JavaTimeModule();
+        JavaTimeModule javaTimeModule = new JavaTimeModule();
+        javaTimeModule.addDeserializer(LocalDateTime.class, new CustomLocalDateTimeDeserializer());
```
sdegroot, line 56: It feels like this more of a workaround for a different problem. Why use a local date time if you receive a zoned date time? 

Would it be an idea to optionally enable (or disable) this? Not sure if everyone wants this behaviour for all LocalDateTime variables by default. You really are removing information after all.

Even though 10:00 UTC+2 is the same moment as 08:00 UTC, it still is a different value.

marijnritense (approved):

sdegroot (commented):
Had eerder al een comment geplaatst maar blijkbaar moet je deze tegenwoordig submitten in GitHub hehe

core/src/main/java/com/ritense/valtimo/config/JacksonConfiguration.java
haiko-ritense: agree, you are removing the zone info

PR #1233, opened 2024-02-23
Added support for RabbitMQ exchanges

theo-ritense:


ThomasMinkeRitense:
Shouldn't this also be done in 11.2 since we introduced the outbox in that version?

theo-ritense:
@ThomasMinkeRitense No, this is a new feature.

Klaas-Ritense (approved):

PR #1235, opened 2024-02-23
fix: start-hersteltermijn

Klaas-Ritense:


zgw/zaken-api/src/main/kotlin/com/ritense/zakenapi/ZakenApiPlugin.kt
```diff
@@ -356,22 +356,21 @@ class ZakenApiPlugin(
                 maxDurationInDays = maxDurationInDays
             )
 
-            val existingHerseltermijn = zaakHersteltermijnRepository.findByZaakUrlAndEndDateIsNull(zaakUrl)
-            check(existingHerseltermijn == null || existingHerseltermijn != hersteltermijn) { "Hersteltermijn already exists for zaak '$zaakUrl'. " }
+            require(zaakHersteltermijnRepository.findByZaakUrlAndEndDateIsNull(zaakUrl) != null) {
```
ThomasMinkeRitense: Shouldn't this be == null?

Klaas-Ritense: Yes, I've added a test for this

zgw/zaken-api/src/main/kotlin/com/ritense/zakenapi/ZakenApiPlugin.kt
```diff
@@ -356,22 +356,21 @@ class ZakenApiPlugin(
                 maxDurationInDays = maxDurationInDays
             )
 
-            val existingHerseltermijn = zaakHersteltermijnRepository.findByZaakUrlAndEndDateIsNull(zaakUrl)
-            check(existingHerseltermijn == null || existingHerseltermijn != hersteltermijn) { "Hersteltermijn already exists for zaak '$zaakUrl'. " }
+            require(zaakHersteltermijnRepository.findByZaakUrlAndEndDateIsNull(zaakUrl) == null) {
```
theo-ritense, line 361: If we want to be on the safe side to prevent data corruption, PostgreSQL supports partial (conditional) constraints on indexes:
https://www.postgresql.org/docs/current/indexes-partial.html#INDEXES-PARTIAL-EX3

Klaas-Ritense: Not sure what you're asking here

theo-ritense: The code is checking if a hersteltermijn already exists, but that could lead to the same type of concurrency/multi-instance issues we faced for documents, which might lead to multiple active hersteltermijnen.

To solve this, you could add a 'unique index' to Postgres like this:
`CREATE UNIQUE INDEX uq_hersteltermijn ON zaak_hersteltermijn (zaak_url, end_date)
    WHERE end_date IS NULL;`

MySQL does not support this unfortunately, but luckily the check in the code alone will probably work fine > 99% of the time.

Klaas-Ritense: Ive added the index

zgw/zaken-api/src/main/resources/config/liquibase/changelog/20240214-create-table-zaak-hersteltermijn.xml
```diff
@@ -37,6 +37,13 @@
         </createTable>
     </changeSet>
 
+    <changeSet id="2" author="Ritense">
+        <createIndex indexName="idx_zaak_hersteltermijn_zaak_url_end_date" tableName="zaak_hersteltermijn" unique="true">
+            <column name="zaak_url"/>
+            <column name="end_date"/>
+        </createIndex>
```
theo-ritense: Your version would not allow 2 hersteltermijnen to be ended on the same day. The following does, because the index is only applied to rows where end_date is null:
```suggestion
         <preConditions onFail="MARK_RAN">
            <dbms type="postgresql"/>
        </preConditions>
        <sql>
            CREATE UNIQUE INDEX idx_zaak_hersteltermijn_zaak_url_end_date ON zaak_hersteltermijn (zaak_url, end_date) WHERE end_date IS NULL;
        </sql>
```

Klaas-Ritense: Changed

ivo-ritense (approved):

PR #1245, opened 2024-02-27
Extended VerzoekPlugins specified copy strategy

Ayub-ritense:
Extended VerzoekPlugins specified copy strategy to map the verzoek data to a process variable when the target prefix is 'pv'

### DIFF
--- a/process-document/src/main/kotlin/com/ritense/processdocument/service/ProcessDocumentsService.kt
+++ b/process-document/src/main/kotlin/com/ritense/processdocument/service/ProcessDocumentsService.kt
@@ -22,6 +22,7 @@ import com.ritense.document.domain.impl.JsonSchemaDocumentId
 import com.ritense.document.exception.DocumentNotFoundException
 import com.ritense.document.service.DocumentService
 import com.ritense.valtimo.service.CamundaProcessService
+import org.camunda.bpm.engine.delegate.DelegateExecution
 import java.util.UUID
 
 
@@ -30,6 +31,10 @@ class ProcessDocumentsService(
     private val camundaProcessService: CamundaProcessService,
     private val associationService: ProcessDocumentAssociationService
 ) {
+    fun startProcessByProcessDefinitionKeyWithExecution(processDefinitionKey: String, execution: DelegateExecution) {
+        startProcessByProcessDefinitionKey(processDefinitionKey, execution.businessKey, execution.variables)
+    }
+
     //TODO: Determine what to with this
     fun startProcessByProcessDefinitionKey(processDefinitionKey: String, businessKey: String) {
         startProcessByProcessDefinitionKey(processDefinitionKey, businessKey, null)
--- a/process-document/src/main/kotlin/com/ritense/processdocument/service/ProcessDocumentsService.kt
+++ b/process-document/src/main/kotlin/com/ritense/processdocument/service/ProcessDocumentsService.kt
@@ -22,7 +22,6 @@ import com.ritense.document.domain.impl.JsonSchemaDocumentId
 import com.ritense.document.exception.DocumentNotFoundException
 import com.ritense.document.service.DocumentService
 import com.ritense.valtimo.service.CamundaProcessService
-import org.camunda.bpm.engine.delegate.DelegateExecution
 import java.util.UUID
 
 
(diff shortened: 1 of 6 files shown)


theo-ritense (commented):
I feel like defining the "doc:" and "pv:" prefixes over and over again is a bit messy. Can't we use `DocumentJsonValueResolverFactory.PREFIX` and `ProcessVariableValueResolverFactory.PREFIX`?

Ayub-ritense:
> I feel like defining the "doc:" and "pv:" prefixes over and over again is a bit messy. Can't we use `DocumentJsonValueResolverFactory.PREFIX` and `ProcessVariableValueResolverFactory.PREFIX`?

Indeed we can use that instead. I will update it thank you!
edit: feedback processed.

process-document/src/main/kotlin/com/ritense/processdocument/service/ProcessDocumentsService.kt
```diff
@@ -30,6 +31,10 @@ class ProcessDocumentsService(
     private val camundaProcessService: CamundaProcessService,
     private val associationService: ProcessDocumentAssociationService
 ) {
+    fun startProcessByProcessDefinitionKeyWithExecution(processDefinitionKey: String, execution: DelegateExecution) {
```
marijnritense: I think if there's variables, they should just be provided, This can be seen on line 44, so I think this method should be removed.

zgw/verzoek/src/main/kotlin/com/ritense/verzoek/VerzoekPluginEventListener.kt
```diff
@@ -38,10 +38,14 @@ import com.ritense.processdocument.service.ProcessDocumentService
...
 import com.ritense.verzoek.domain.CopyStrategy
 import com.ritense.verzoek.domain.VerzoekProperties
+import com.ritense.processdocument.resolver.DocumentJsonValueResolverFactory.Companion.PREFIX as DOC_PREFIX
+import com.ritense.valueresolver.ProcessVariableValueResolverFactory.Companion.PREFIX as PV_PREFIX
 import mu.KotlinLogging
 import org.springframework.context.event.EventListener
+import org.springframework.transaction.annotation.Transactional
 import java.net.URI
 
+@Transactional
```
marijnritense, line 48: You're missing a dependency for this

zgw/verzoek/src/main/kotlin/com/ritense/verzoek/VerzoekPluginEventListener.kt
```diff
@@ -91,6 +95,28 @@ open class VerzoekPluginEventListener(
...
+                    property.mapping?.forEach {
+                        if (it.target.startsWith(PV_PREFIX)) {
+                            try {
+                                val key = it.target.substringAfter(delimiter = "/")
+                                val pointer = "/data/${it.source.substringAfter(delimiter = "/")}"
+
+                                val value = getJsonValueWithPointer(verzoekObjectData, pointer)
+                                verzoekVariables[key] = value
+
+                            } catch (e: Exception) {
```
marijnritense: This is kind of a catch all. Is there any information that should be relayed to improve debugging here?

zgw/verzoek/src/main/kotlin/com/ritense/verzoek/VerzoekPluginEventListener.kt
```diff
@@ -195,6 +223,14 @@ open class VerzoekPluginEventListener(
         }
     }
 
+    private fun getJsonValueWithPointer(jsonNode: JsonNode, pointer: String): String? {
+        val valueNode = jsonNode.at(pointer)
+        if (valueNode.isMissingNode) {
+            throw NoSuchElementException("Path does not exist in the JSON: $pointer")
+        }
+        return valueNode.asText()
```
Klaas-Ritense: - A missing node should be saved as `null`. This is also how its done for the `doc:` prefix
- `asText()` will save booleans and numbers as a String 

```suggestion
        return if (jsonNode == null || jsonNode.isMissingNode || jsonNode.isNull) {
            null
        } else if (jsonNode.isValueNode || jsonNode.isArray || jsonNode.isObject) {
            objectMapper.treeToValue(jsonNode, Object::class.java)
        } else {
            jsonNode.asText()
        }
```

Klaas-Ritense (approved):

PR #1249, opened 2024-02-29
#103006 - [BE] Architecture to store custom task columns

IvarKoreman-Ritense:
https://ritense.tpondemand.com/entity/103006-be-architecture-to-store-custom-task


case/src/main/kotlin/com/ritense/case/configuration/CaseAutoConfiguration.kt
```diff
@@ -93,6 +96,14 @@ class CaseAutoConfiguration {
         return CaseInstanceResource(service)
     }
 
+    @Bean
+    @ConditionalOnMissingBean(name = ["taskListResource"]) // because integration tests fail to initialise in portaaltaak
```
Klaas-Ritense: In the future we want to use the `@ConditionalOnMissingBean(MyService::class)` on all beans. So not based on `name` but on `class`. Because in the future:
- All services in Valtimo will get an interface
- Interfaces can be overriden thanks to the `@ConditionalOnMissingBean(MyService::class)`

case/src/main/kotlin/com/ritense/case/security/config/CaseHttpSecurityConfigurer.kt
```diff
@@ -42,10 +42,14 @@ class CaseHttpSecurityConfigurer : HttpSecurityConfigurer {
...
-                    .requestMatchers(antMatcher(GET, MANAGEMENT_LIST_COLUMN_URL)).hasAuthority(ADMIN)
-                    .requestMatchers(antMatcher(POST, MANAGEMENT_LIST_COLUMN_URL)).hasAuthority(ADMIN)
-                    .requestMatchers(antMatcher(PUT, MANAGEMENT_LIST_COLUMN_URL)).hasAuthority(ADMIN)
-                    .requestMatchers(antMatcher(DELETE, "$MANAGEMENT_LIST_COLUMN_URL/{columnKey}")).hasAuthority(ADMIN)
+                    .requestMatchers(antMatcher(GET, MANAGEMENT_CASE_LIST_COLUMN_URL)).hasAuthority(ADMIN)
+                    .requestMatchers(antMatcher(POST, MANAGEMENT_CASE_LIST_COLUMN_URL)).hasAuthority(ADMIN)
+                    .requestMatchers(antMatcher(PUT, MANAGEMENT_CASE_LIST_COLUMN_URL)).hasAuthority(ADMIN)
+                    .requestMatchers(antMatcher(DELETE, "$MANAGEMENT_CASE_LIST_COLUMN_URL/{columnKey}")).hasAuthority(ADMIN)
+                    .requestMatchers(antMatcher(GET, MANAGEMENT_TASK_LIST_COLUMN_URL)).hasAuthority(ADMIN)
+                    .requestMatchers(antMatcher(POST, "$MANAGEMENT_TASK_LIST_COLUMN_URL/{columnKey}")).hasAuthority(ADMIN)
```
Klaas-Ritense: A `POST` isn't needed according to the story

case/src/main/kotlin/com/ritense/case/web/rest/TaskListResource.kt
Klaas-Ritense, line 1: Missing endpoint `PUT /management/v1/case/{caseDefinitionName}/task-list-column `

IvarKoreman-Ritense:
Processed the feedback. I've changed the PUT to change the column order to a POST, since I don't see how we can make this idempotent. (Unless we submit all columns like we do for the Case list columns, but that also doesn't seem like the best approach.)

@Klaas-Ritense

Klaas-Ritense (dismissed):

case/src/main/kotlin/com/ritense/case/domain/TaskListColumn.kt
```diff
@@ -0,0 +1,47 @@
...
+import jakarta.persistence.EnumType
+import jakarta.persistence.Enumerated
+import jakarta.persistence.Table
+import org.hibernate.annotations.Type
+
+@Entity
+@Table(name = "task_list_column")
+data class TaskListColumn(
+    @EmbeddedId
+    val id: CaseListColumnId,
```
marijnritense: Can you change this to a TaskListColumnId?

Klaas-Ritense (approved):

PR #1261, opened 2024-03-06
Reverted permission check because it's a breaking change

ThomasMinkeRitense:


marijnritense:
Don't forget there's also code in `CamundaProcessJsonSchemaDocumentAssociationService` that also has to be reverted. We do the same permission check there.

ivo-ritense (approved):

PR #1267, opened 2024-03-08
feat: autodeployment document list columns

mbritense:
Closes https://ritense.tpondemand.com/entity/104923-be-autodeployment-document-list-columns


app/gzac/src/main/resources/config/documenten-api/leningen.zgw-document-list-column.json
```diff
@@ -0,0 +1,16 @@
+{
+    "changesetId": "leningen.zgw-document-list-columns-v1",
+    "case-definitions": [
+        {
+            "key": "leningen",
+            "columns": [
```
marijnritense, line 6: There's code in the `DocumentenApiColumnDeploymentService` that should be updated with the introduction of autodeployment. 

Currently, when a document definition is deployed, the default DocumentenApiColumns are set for that document definition. This should only happen, however, if there's no applicable `zgw-document-list-column` files. This requires looking into the contents of the file.

Klaas-Ritense (approved):

PR #1284, opened 2024-03-20
#104289 Weird keys can be used as a key for internalStatuses

Klaas-Ritense:


contract/src/main/java/com/ritense/valtimo/contract/web/rest/error/ExceptionTranslator.java
```diff
@@ -65,9 +65,11 @@ public ResponseEntity<Problem> process(@Nullable ResponseEntity<Problem> entity,
         }
 
         final String msg = "message";
-        if (problem instanceof ConstraintViolationProblem) {
+        if (problem instanceof ConstraintViolationProblem constraintViolationProblem) {
             builder
-                .with("violations", ((ConstraintViolationProblem) problem).getViolations())
+                .with("errors", constraintViolationProblem.getViolations().stream()
```
ivo-ritense, line 70: Not sure i'm happy with this. This is a breaking change that will impact any @valid we use in the codebase. Do we want to do this?

Klaas-Ritense: This code exclusively influences the HTTP response generated by `@Valid`. Currently, the application fails to correctly display this error. In instances where there are multiple violations in a single response, the frontend anticipates them to be in the errors field.

We could still keep the `violations` field to make it non-breaking. But I haven't seen anyone use it.

theo-ritense: This is a major after all, so I'm not too worried about a 'breaking' change even if that's true. 

I'm more worried that it might be leaking information, but that's a general concern I have with the inner workings of the hardeningservice and its configuration.

PS: I think this PR is fine.

theo-ritense (approved):

PR #1297, opened 2024-04-02
#101331 Zaakdetail sync through plugins

Klaas-Ritense:


zgw/object-management/src/main/kotlin/com/ritense/objectmanagement/autoconfigure/ObjectManagementAutoConfiguration.kt
```diff
@@ -69,6 +70,7 @@ class ObjectManagementAutoConfiguration {
         )
     }
 
+    @Primary
```
ivo-ritense: I don't like having primary beans in the valtimo because you can't easily override them. Why is this required here?

Klaas-Ritense: Changed

zgw/object-management/src/main/kotlin/com/ritense/objectmanagement/security/config/ObjectManagementHttpSecurityConfigurer.kt
```diff
@@ -38,6 +38,8 @@ class ObjectManagementHttpSecurityConfigurer : HttpSecurityConfigurer {
                     .requestMatchers(antMatcher(DELETE, "$CONFIGURATION_URL/{id}")).hasAuthority(ADMIN)
                     .requestMatchers(antMatcher(GET, "$CONFIGURATION_URL/{id}/object")).authenticated()
                     .requestMatchers(antMatcher(POST, "$CONFIGURATION_URL/{id}/object")).authenticated()
+
+                    .requestMatchers(antMatcher(GET, CONFIGURATION_MANAGEMENT_URL)).authenticated()
```
theo-ritense: This should be ADMIN only

zgw/objecten-api/src/main/kotlin/com/ritense/objectenapi/client/ObjectenApiClient.kt
```diff
@@ -260,14 +260,25 @@ class ObjectenApiClient(
         objectUrl: URI,
         objectRequest: ObjectRequest
     ): ObjectWrapper {
+        val objectRequestCorrectedHost = if (objectRequest.type.host == "localhost") {
```
theo-ritense, line 263: We should make a filter for this that is only applied when running in development

zgw/object-management/src/main/kotlin/com/ritense/objectmanagement/security/config/ObjectManagementHttpSecurityConfigurer.kt
Klaas-Ritense: Fixed

theo-ritense (approved):

PR #1308, opened 2024-04-12
Use processdefinitionkey as filename as keys with more than 25 charac…

ThomasMinkeRitense:
…ters are excluded from the id


theo-ritense (dismissed):
Please fix the tests. Otherwise it looks ok.

ThomasMinkeRitense:
@theo-ritense please check again

marijnritense (approved):

PR #1313, opened 2024-04-17
#109221 assigning a case to a different user does not work

Klaas-Ritense:


theo-ritense (commented):
We're going back to assigning by user id?

document/src/main/java/com/ritense/document/service/impl/JsonSchemaDocumentService.java
```diff
@@ -582,7 +582,7 @@ public void assignUserToDocument(UUID documentId, String assigneeId) {
                 new DelegateUserEntityAuthorizationRequest<>(
                     JsonSchemaDocument.class,
                     ASSIGNABLE,
-                    assignee.getEmail(),
```
Klaas-Ritense, line 585: @theo-ritense 
> We're going back to assigning by user id?

We're in the process of switching from email to user-id. See ticket: #68171

ivo-ritense (approved):

PR #1314, opened 2024-04-18
Feature/109667-setup-form-view-model-module

### LINKED ISSUE #1350: ViewModels and Commands
### Probleemstelling:
Data structureren zijn vaak te complex waardoor het mappen ervan bemoeilijkt wordt.
Delen van het formulier mappen niet direct op een onderliggende structuur en andersom
Data automatisch mappen kan ingewikkeld worden.

Risico’s/tekortkomingen:
- null mappings worden genegeerd
- waardes worden overschreven
- omgang met arrays
- complexe validatie niet mogelijk binnen Valtimo Out of the box.

Data veranderingen leiden to formulieren die op runtime falen.

### Oplossing
concepts-form-view-model drawio

### Ontkoppel formulier definitie van data definitie

1. ViewModel: fungeert als een bemiddelaar tussen de view (gebruikersinterface-elementen) en het model (gegevens). Versimpelt beheer en presentatie
2. Commands: bevordert een duidelijke scheiding van belangen en een modulaire architectuur, wat het gemakkelijker maakt om complexe softwaresystemen te ontwikkelen en te evolueren.
3. Zorg voor consistentie in de hele keten, om verrassingen tijdens de uitvoering te voorkomen. 
Contracten worden gevalideerd elke keer dat ze worden gewijzigd.

Tom-Ritense:
Setup of new Valtimo module called 'form-view-model'


form/src/main/kotlin/com/ritense/form/processlink/FormProcessLinkActivityHandler.kt
```diff
@@ -32,7 +32,7 @@ class FormProcessLinkActivityHandler(
 ) : ProcessLinkActivityHandler<FormTaskOpenResultProperties> {
 
     override fun supports(processLink: ProcessLink): Boolean {
-        return processLink is FormProcessLink
+        return processLink is FormProcessLink && !processLink.viewModelEnabled
```
marijnritense, line 33: The first `ProcessLinkActivityHandler` for which the `supports` method evaluates to true will be used. So you can give the `FormViewModelProcessLinkActivityHandler` a higher priority and it will be used instead.

form-view-model/src/main/kotlin/com/ritense/formviewmodel/viewmodel/ViewModelLoader.kt
```diff
@@ -0,0 +1,35 @@
...
+
+@Transactional
+interface ViewModelLoader<T : ViewModel> {
+
+    fun load(task: CamundaTask): T
+
+    fun supports(formName: String) = getFormName() == formName
+
+    @Suppress("UNCHECKED_CAST")
+    fun getViewModelType(): KClass<T> = this::class.supertypes.first().arguments.first().type!!.classifier as KClass<T>
```
marijnritense: Please do firstOrNull and handle the null gracefully.

form-view-model/src/main/kotlin/com/ritense/formviewmodel/commandhandling/Command.kt
```diff
@@ -0,0 +1,19 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" basis,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.ritense.formviewmodel.commandhandling
+
+interface Command<out T>
```
marijnritense: Not going to comment this everywhere, but you're adding a lot of code meant to make life easier in a generic way.  But it's put in a module that is very specific. I'd say either remove all of this (and refactor the rest of the code) or move it to a more generic module.
Alternatively: Maybe choose a different name for this. Command feels very generic. Maybe rename to `FormViewModelCommand`?

form-view-model/src/main/kotlin/com/ritense/formviewmodel/web/rest/error/FormViewModelModuleExceptionTranslator.kt
```diff
@@ -0,0 +1,39 @@
...
+class FormViewModelModuleExceptionTranslator : ProblemHandling {
+
+    @ExceptionHandler(FormException::class)
+    fun handleFormException(
+        ex: FormException,
+        request: NativeWebRequest
+    ): ResponseEntity<FormError> {
+        return ResponseEntity
+            .badRequest()
+            .body(FormError(error = ex.message!!, component = ex.component))
```
marijnritense: Only going to put this down once, but I think this needs to be fixed everywhere. Why are you using double exclamation marks? There are graceful ways of handling nullability. You could now get an exception when trying to handle an exception.

marijnritense: For this specific exception as well: Maybe the FormError should take the exception and distill the message and component from there?

form/src/main/resources/config/liquibase/changelog/20230330-add-form-process-link.xml
```diff
@@ -28,4 +28,13 @@
         </addColumn>
     </changeSet>
 
+    <changeSet id="2" author="Tom Bokma">
```
marijnritense: Please put this in a new changelog with the current date in the filename.

form-view-model/src/main/kotlin/com/ritense/formviewmodel/commandhandling/decorator/DecoratorBuilder.kt
```diff
@@ -0,0 +1,46 @@
...
+ * limitations under the License.
+ */
+
+package com.ritense.formviewmodel.commandhandling.decorator
+
+import com.ritense.formviewmodel.commandhandling.Command
+import com.ritense.formviewmodel.commandhandling.CommandHandler
+import mu.KotlinLogging
+
+class DecoratorBuilder<T>(private val handler: CommandHandler<Command<T>, T>) {
```
marijnritense: Is this needed for this feature? I understand it's useful, but I don't think anyone is asking for this per say. Not sure what to think with this.

form-view-model/src/main/kotlin/com/ritense/formviewmodel/json/MapperSingleton.kt
```diff
@@ -0,0 +1,51 @@
...
+
+package com.ritense.formviewmodel.json
+
+import com.fasterxml.jackson.databind.ObjectMapper
+import com.fasterxml.jackson.datatype.jsr310.ser.LocalDateTimeSerializer
+import org.springframework.boot.autoconfigure.jackson.Jackson2ObjectMapperBuilderCustomizer
+import org.springframework.http.converter.json.Jackson2ObjectMapperBuilder
+import java.time.format.DateTimeFormatter
+
+object MapperSingleton {
```
marijnritense: We already include this in Contract.

form-view-model/src/main/kotlin/com/ritense/formviewmodel/web/rest/FormViewModelResource.kt
```diff
@@ -0,0 +1,105 @@
...
+import org.springframework.web.bind.annotation.PostMapping
+import org.springframework.web.bind.annotation.RequestBody
+import org.springframework.web.bind.annotation.RequestMapping
+import org.springframework.web.bind.annotation.RequestParam
+import org.springframework.web.bind.annotation.RestController
+
+@RestController
+@SkipComponentScan
+@RequestMapping("/api/v1/form/view-model", produces = [APPLICATION_JSON_UTF8_VALUE])
+class FormViewModelResource(
```
marijnritense, line 38: There's a bit of logic in this resource. Please move service-level logic to a service, e.g. authorization checks.

form-view-model/src/test/kotlin/com/ritense/formviewmodel/commandhandling/CommandDispatcherTest.kt
```diff
@@ -0,0 +1,60 @@
...
+import com.ritense.formviewmodel.commandhandling.CommandDispatcher
+import com.ritense.formviewmodel.commandhandling.CommandHandler
+import org.assertj.core.api.Assertions.assertThat
+import org.junit.jupiter.api.BeforeEach
+import org.junit.jupiter.api.Test
+import org.junit.jupiter.api.assertDoesNotThrow
+import org.mockito.kotlin.spy
+import org.mockito.kotlin.verify
+
+internal class CommandDispatcherTest : BaseTest() {
```
marijnritense: What if a handler fails to execute? Would like a test case for that as well.

form-view-model/src/test/kotlin/com/ritense/formviewmodel/commandhandling/decorator/DecoratorBuilderTest.kt
```diff
@@ -0,0 +1,28 @@
...
+    @Test
+    fun `should build with default decorators`() {
+        val commandHandler: CommandHandler<Command<Unit>, Unit> = mock()
+        val builder = spy(DecoratorBuilder(commandHandler))
+        val commandHandlerDecorated = builder
+            .decorateWithDefaults()
+            .build()
+
+        assertThat(commandHandlerDecorated).isNotNull
+        verify(builder).decorateWithDefaults()
```
marijnritense: Can we also have tests that verify the individual decorators work as they should?

form-view-model/src/test/kotlin/com/ritense/formviewmodel/event/FormViewModelSubmissionHandlerFactoryTest.kt
```diff
@@ -0,0 +1,24 @@
...
+
+    @BeforeEach
+    fun setUp() {
+        formViewModelSubmissionHandlerFactory = FormViewModelSubmissionHandlerFactory(
+            listOf(TestSubmissionHandler())
+        )
+    }
+
+    @Test
+    fun `should create submission handler`() {
```
marijnritense: Another test to verify what happens when no handler can be found?

form-view-model/src/test/kotlin/com/ritense/formviewmodel/service/FormViewModelSubmissionServiceTest.kt
```diff
@@ -0,0 +1,75 @@
...
+            formName = "test",
+            submission = submission,
+            task = camundaTask
+        )
+        verify(commandDispatcher).dispatch(any<ExampleCommand>())
+        verify(camundaTaskService).complete(camundaTask.id)
+    }
+
+    @Test
+    fun `should not handle submission`() {
```
marijnritense: Can you make this a bit more descriptive? E.g. `should not handle submission when exception is thrown`

form-view-model/src/test/kotlin/com/ritense/formviewmodel/validation/OnStartUpViewModelValidatorTest.kt
```diff
@@ -0,0 +1,146 @@
...
+        val testViewModelLoader = TestViewModelLoader()
+        val missingFields = onStartUpViewModelValidator.validateViewModel(
+            testViewModelLoader,
+            formDefinitionOf("user-task-1")
+        )
+        assertThat(missingFields).isEmpty()
+    }
+
+    @Test
+    fun `should not be valid ViewModel`() {
```
marijnritense: Can you make this a bit more descriptive? Goes for every test here.

marijnritense (dismissed):
Left some comments on individual files/lines.

Also missing some integration tests to verify a viewmodel can be loaded, and a submission can be saved.

form-view-model/src/main/kotlin/com/ritense/formviewmodel/web/rest/error/FormViewModelModuleExceptionTranslator.kt
sirolf2009: I think it's nicer if the exception translator builds up the FormError. The FormError is just a DTO and I'd rather not have any logic in there

form-view-model/src/test/kotlin/com/ritense/formviewmodel/commandhandling/CommandDispatcherTest.kt
sirolf2009: It is up to whoever executes a command to catch the exception and somehow deal with it. For form view model, the errors are caught in the exception translator. I don't know if a test that says "assert that an exception is thrown when an exception is thrown" really makes sense

marijnritense: You would be testing whether the method performs any transformation to an exception. When doing unit tests, it's meaningful to go through every path in a method. As it stands, someone could remove the thrown exception at the end of that `catch`, which could be bad for underlying code.

form-view-model/src/main/kotlin/com/ritense/formviewmodel/commandhandling/decorator/DecoratorBuilder.kt
sirolf2009: We copied this command code from Allnex which included these decorators that were really helpful to us. We will have a discussion in team Galaxy if we'd like to make these commands a separate module

form-view-model/src/main/kotlin/com/ritense/formviewmodel/web/rest/FormViewModelResource.kt
Tom-Ritense: I Would like to challenge this observation. Looking at the code of Valtimo I see everywhere auth checks being made. I don't want to mix this checking into a ViewLoader factory for example. Reasoning is that this is just something different and joining these concepts seems off. So then you could argue making a dedicated auth service for checks. A bit better approach maybe but then the benefit is kinda low for my taste.
I lean towards accepting code below being used in a resource file. Any thoughts?
 ```
   val task = taskInstanceId.let {
            camundaTaskService.findTaskById(it).also { foundTask ->
                authorizationService.requirePermission(
                    EntityAuthorizationRequest(CamundaTask::class.java, VIEW, foundTask)
                )
            }
        }
```

File: form-view-model/src/main/kotlin/com/ritense/formviewmodel/web/rest/error/FormViewModelModuleExceptionTranslator.kt (hunk 4 above)
Comment (marijnritense) [reply form-view-model/src/main/kotlin/com/ritense/formviewmodel/web/rest/error/FormViewModelModuleExceptionTranslator.kt]: Sure! Could you then at least move the `"Unknown Form Error"` message into the constructor for the FormError? You would want that part at least to be consistent, which you can only really enforce on a constructor level.

File: form-view-model/src/main/kotlin/com/ritense/formviewmodel/web/rest/error/FormViewModelModuleExceptionTranslator.kt [hunk 14]
```diff
@@ -0,0 +1,39 @@
...
+    }
+
+    @ExceptionHandler(BusinessException::class)
+    fun handleBusinessException(
+        ex: BusinessException,
+        request: NativeWebRequest
+    ): ResponseEntity<BusinessRuleError> {
+        return ResponseEntity
+            .badRequest()
+            .body(BusinessRuleError(error = ex.message ?: "Unknown Business Rule Error"))
```
marijnritense: Same here, please move the `"Unknown Business Rule Error"` to the constructor.

form-view-model/src/main/kotlin/com/ritense/formviewmodel/web/rest/error/FormViewModelModuleExceptionTranslator.kt
sirolf2009: That is a little bit difficult since `String` and `String?` have the same signature. It's also still putting in logic (`?:`) in a DTO which I'm not a fan of. 

What if I create a companion method instead to build the object from an Exception instead? The logic won't be completely part of the DTO that way, but they are still not very separated either

form-view-model/src/main/kotlin/com/ritense/formviewmodel/service/FormViewModelService.kt
```diff
@@ -20,63 +20,86 @@ import com.fasterxml.jackson.databind.ObjectMapper
...
 ) {
-    fun getFormViewModel(
+
+    fun getStartFormViewModel(
+        formName: String,
+        processDefinitionId: String
+    ): ViewModel? {
+        require(
+            authorizationService.hasPermission(
+                RelatedEntityAuthorizationRequest(
```
marijnritense: I don't feel confident this is the right permission. You can make the argument being able to see a form requires `VIEW` permissions, in which case the resource this should apply to changes as well.

Tom-Ritense: Let switch to the VIEW then

form-view-model/src/main/kotlin/com/ritense/formviewmodel/web/rest/error/FormViewModelModuleExceptionTranslator.kt
sirolf2009: I ended up doing this through a builder, that way the `Unknown Form Error` can still be consistent but without any logic in the DTO

command-handling/src/main/kotlin/com/ritense/commandhandling/SpringContextHelper.kt
```diff
@@ -0,0 +1,49 @@
...
+ * limitations under the License.
+ */
+
+package com.ritense.commandhandling
+
+import mu.KotlinLogging
+import org.springframework.context.ApplicationContext
+import org.springframework.context.ApplicationContextAware
+
+class SpringContextHelper : ApplicationContextAware {
```
theo-ritense, line 23: This class seems to be unused. Why is it there?

form-view-model/src/main/kotlin/com/ritense/formviewmodel/SpringContextHelper.kt
```diff
@@ -0,0 +1,49 @@
...
+ * limitations under the License.
+ */
+
+package com.ritense.formviewmodel
+
+import mu.KotlinLogging
+import org.springframework.context.ApplicationContext
+import org.springframework.context.ApplicationContextAware
+
+class SpringContextHelper : ApplicationContextAware {
```
theo-ritense: This class seems to be unused. Why is it there?

form-view-model/src/main/kotlin/com/ritense/formviewmodel/service/FormViewModelService.kt
```diff
@@ -0,0 +1,92 @@
...
+import com.ritense.authorization.request.EntityAuthorizationRequest
+import com.ritense.formviewmodel.viewmodel.ViewModel
+import com.ritense.formviewmodel.viewmodel.ViewModelLoaderFactory
+import com.ritense.valtimo.camunda.authorization.CamundaTaskActionProvider.Companion.VIEW
+import com.ritense.valtimo.camunda.domain.CamundaTask
+import com.ritense.valtimo.service.CamundaTaskService
+import kotlin.reflect.KClass
+
+class FormViewModelService(
+    val objectMapper: ObjectMapper,
```
theo-ritense: private

form-view-model/src/main/kotlin/com/ritense/formviewmodel/web/rest/dto/BusinessRuleError.kt
```diff
@@ -0,0 +1,39 @@
...
+ * limitations under the License.
+ */
+
+package com.ritense.formviewmodel.web.rest.dto
+
+data class BusinessRuleError private constructor (
+    val error: String
+) {
+
+    class Builder {
```
theo-ritense: I don't see any reason to use Builders with Kotlin. The only use case I see here could be solved by a default value for `error`.

form-view-model/src/test/kotlin/com/ritense/formviewmodel/service/FormViewModelServiceTest.kt
```diff
@@ -0,0 +1,125 @@
...
+    private lateinit var authorizationService: AuthorizationService
+    private lateinit var formViewModelSubmissionService: FormViewModelSubmissionService
+    private lateinit var camundaTask: CamundaTask
+    private lateinit var processAuthorizationService: ProcessAuthorizationService
+    private lateinit var formViewModelService: FormViewModelService
+
+    @BeforeEach
+    fun setUp() {
+        objectMapper = MapperSingleton.get()
+        viewModelLoaderFactory = mock()
```
theo-ritense, line 36: Most of this setup can be eliminated by using constructor injection and `@Mock`.

form-view-model/src/test/kotlin/com/ritense/formviewmodel/validation/OnStartUpViewModelValidatorTest.kt
```diff
@@ -0,0 +1,150 @@
...
+    @InjectMocks
+    private lateinit var onStartUpViewModelValidator: OnStartUpViewModelValidator
+    private lateinit var viewModelLoaders: List<ViewModelLoader<*>>
+    private lateinit var viewModelLoader: ViewModelLoader<ViewModel>
+    private lateinit var formViewModelStartFormSubmissionHandlerFactory: FormViewModelStartFormSubmissionHandlerFactory
+    private lateinit var formViewModelUserTaskSubmissionHandlerFactory: FormViewModelUserTaskSubmissionHandlerFactory
+
+    @BeforeEach
+    fun setUp() {
+        formIoFormDefinitionService = mock()
```
theo-ritense, line 36: This is some limbo between using `@Mock` vs `mock()`. I prefer using constructor injection so that we also get rid of the `lateinit var` stuff.

theo-ritense (commented):
I'm generally not a fan of using commands, since it makes navigating through and understanding of the code harder.

Some code seems to be unused, like the implemented commands and decorators. But that might just be because I looked at it from the code perspective, and not functionally (yet).

form-view-model/src/main/kotlin/com/ritense/formviewmodel/SpringContextHelper.kt
Tom-Ritense: It is used when I look at it in intelliJ 

```
fun <T> dispatchCommand(command: Command<T>) =
    SpringContextHelper.getBean(CommandDispatcher::class.java).dispatch(command)
```

form-view-model/src/main/kotlin/com/ritense/formviewmodel/service/FormViewModelService.kt
Tom-Ritense: This function needs it to be public 

```
inline fun <reified T : ViewModel> parseViewModel(
        submission: ObjectNode,
        viewModelType: KClass<out T>
    ): ViewModel {
        // When a field is not present in the ViewModel what then? A: it's ignored
        return objectMapper.convertValue(submission, viewModelType.java)
    }
```

form-view-model/src/main/kotlin/com/ritense/formviewmodel/web/rest/dto/BusinessRuleError.kt
Tom-Ritense: @sirolf2009 can you explain why you used this approach?

form-view-model/src/test/kotlin/com/ritense/formviewmodel/service/FormViewModelServiceTest.kt
Tom-Ritense: Yes, mock() is the equivalent of @Mock. :)

form-view-model/src/test/kotlin/com/ritense/formviewmodel/validation/OnStartUpViewModelValidatorTest.kt
Tom-Ritense: This is an opinion and the nature of open source projects should not be to have 1 view imo. These are minor things imo. To argue for mock() is that @InjectMocks is kinda magic. So I rather use mock() over @Mock. 
Contribution guidelines should then be more clear? what do you think?.

form-view-model/src/main/kotlin/com/ritense/formviewmodel/web/rest/dto/BusinessRuleError.kt
sirolf2009: It's to enforce the default "unknown" message. A default value in the constructor doesn't quite work either, because java returns this message as a `String?`, so adding a default value in the constructor will just give an error that `String? cannot be cast to String`. Somewhere, a `?:` check needs to be done and to keep that consistent, I set up this builder.

form-view-model/src/test/kotlin/com/ritense/formviewmodel/validation/OnStartUpViewModelValidatorTest.kt
theo-ritense: This specific comment was about the `parseViewModel` being annotated with `@Mock`, and then set again in the `setup()`. Same with the `@InjectMocks`. Since `lateinit var` can only be set once, the `@Mock` and `@InjectMocks` annotation clearly aren't working properly. Probably because the class is not annotated with `@ExtendWith(MockitoExtension.class)`. Hence the 'limbo'. 

You are right that the rest of it (constructor injection) is personal preference and I will not decline a PR based on that only.

form-view-model/src/main/kotlin/com/ritense/formviewmodel/SpringContextHelper.kt
theo-ritense: That method is pointing towards the `com.ritense.commandhandling.SpringContextHelper`. Not this one.

form-view-model/src/main/kotlin/com/ritense/formviewmodel/service/FormViewModelService.kt
theo-ritense: That method does not have to be reified. It works just fine when you replace it with this:
```
fun <T : ViewModel> parseViewModel(
        submission: ObjectNode,
        viewModelType: KClass<out T>
    ): ViewModel {
        // When a field is not present in the ViewModel what then? A: it's ignored
        return objectMapper.convertValue(submission, viewModelType.java)
    }
```

The `objectMapper` can then be private

form-view-model/src/main/kotlin/com/ritense/formviewmodel/web/rest/dto/BusinessRuleError.kt
theo-ritense: @sirolf2009 

That would work just as well by doing this, right?:
```
class BusinessRuleError (
    error: String?
) {
    val error: String = error ?: UNKNOWN_BUSINESS_RULE_ERROR

    companion object {
        const val UNKNOWN_BUSINESS_RULE_ERROR = "Unknown Business Rule Error"
    }
}
```

This is more Kotlin-like imho. I'll leave it to you to decide.

form-view-model/src/test/kotlin/com/ritense/formviewmodel/service/FormViewModelServiceTest.kt
theo-ritense: You're mocking me, but the annotation version would cut your setup line size in half, getting the reader faster to the part that is important.

form-view-model/src/main/kotlin/com/ritense/formviewmodel/web/rest/dto/BusinessRuleError.kt
sirolf2009: Huh, didn't know you could do that :)

I'll take your proposal, it's a lot less lines of code

form-view-model/src/test/kotlin/com/ritense/formviewmodel/validation/OnStartUpViewModelValidatorTest.kt
Tom-Ritense: Will remove the annotation

form-view-model/src/main/kotlin/com/ritense/formviewmodel/SpringContextHelper.kt
Tom-Ritense: Will remove

dependencies/valtimo-dependency-versions/build.gradle
```diff
@@ -41,6 +41,7 @@ dependencies {
         api(project(":form"))
         api(project(":form-flow"))
         api(project(":form-flow-valtimo"))
+        api(project(":form-view-model"))
```
theo-ritense, line 44: Please add the `command-handling` module as well.

docker-resources/docker-compose-base-test-mysql.yml
```diff
@@ -65,4 +65,5 @@ services:
 #   url: jdbc:mysql://localhost:3348/outbox-rabbitmq-test
 #   url: jdbc:mysql://localhost:3349/localization-test
 #   url: jdbc:mysql://localhost:3351/zaakdetails-test
+#   url: jdbc:mysql://localhost:3352/form-view-model-test
```
theo-ritense, line 67: A new entry was added for the postgresql version, but not here.

form-view-model/src/main/kotlin/com/ritense/formviewmodel/validation/OnStartUpViewModelValidator.kt
```diff
@@ -0,0 +1,138 @@
...
+            }
+        }
+        return getAllMissingProperties(submissionHandler.getSubmissionType(), formDefinition)
+    }
+
+    private fun getAllMissingProperties(
+        submission: KClass<*>,
+        formDefinition: FormIoFormDefinition
+    ): List<String> {
+        val fieldNames = DataClassPropertiesExtractor.extractProperties(submission)
```
theo-ritense, line 129: This seems you rely heavily on the use of `data` classes. Can this feature also be used with standard Java classes?

command-handling/src/main/kotlin/com/ritense/commandhandling/CommandHandler.kt
```diff
@@ -0,0 +1,32 @@
...
+import kotlin.reflect.KClass
+
+@Transactional
+interface CommandHandler<C : Command<T>, out T> {
+
+    fun execute(command: C) : T
+
+    @Suppress("UNCHECKED_CAST")
+    fun getCommandType() =
+        this::class.supertypes.first().arguments.first().type?.let { it.classifier as KClass<C> }
```
theo-ritense: This only works when `CommandHandler` is the first implemented interface, and is a direct implementation without inheritence.
```suggestion
        this::class.allSupertypes.first { it.classifier == CommandHandler::class }.arguments.first().type?.let { it.classifier as KClass<C> }
```

form-view-model/src/main/kotlin/com/ritense/formviewmodel/submission/FormViewModelStartFormSubmissionHandler.kt
```diff
@@ -0,0 +1,74 @@
...
+     * Retrieves the type of the submission.
+     *
+     * This method uses reflection to determine the type of the submission.
+     *
+     * @return the class type of the submission
+     * @throws IllegalArgumentException if the submission type cannot be resolved
+     */
+    @Suppress("UNCHECKED_CAST")
+    fun getSubmissionType(): KClass<T> =
+        this::class.supertypes.first().arguments.first().type?.let { it.classifier as KClass<T> }
```
theo-ritense: This only works when `FormViewModelStartFormSubmissionHandler` is the first implemented interface, and is a direct implementation without inheritence.
```suggestion
        this::class.allSupertypes.first { it.classifier == FormViewModelStartFormSubmissionHandler::class }.arguments.first().type?.let { it.classifier as KClass<C> }
```

form-view-model/src/main/kotlin/com/ritense/formviewmodel/submission/FormViewModelUserTaskSubmissionHandler.kt
```diff
@@ -0,0 +1,72 @@
...
+     * Retrieves the type of the submission.
+     *
+     * This method uses reflection to determine the type of the submission.
+     *
+     * @return the class type of the submission
+     * @throws IllegalArgumentException if the submission type cannot be resolved
+     */
+    @Suppress("UNCHECKED_CAST")
+    fun getSubmissionType(): KClass<T> =
+        this::class.supertypes.first().arguments.first().type?.let { it.classifier as KClass<T> }
```
theo-ritense: This only works when `FormViewModelUserTaskSubmissionHandler ` is the first implemented interface, and is a direct implementation without inheritence.
```suggestion
        this::class.allSupertypes.first { it.classifier == FormViewModelUserTaskSubmissionHandler::class }.arguments.first().type?.let { it.classifier as KClass<C> }
```

form-view-model/src/main/kotlin/com/ritense/formviewmodel/viewmodel/ViewModelLoader.kt
```diff
@@ -0,0 +1,37 @@
...
+@Transactional
+interface ViewModelLoader<T : ViewModel> {
+
+    fun load(task: CamundaTask? = null): T
+
+    fun supports(formName: String) = getFormName() == formName
+
+    @Suppress("UNCHECKED_CAST")
+    fun getViewModelType(): KClass<T> =
+        this::class.supertypes.first().arguments.first().type?.let { it.classifier as KClass<T> }
```
theo-ritense: This only works when `ViewModelLoader` is the first implemented interface, and is a direct implementation without inheritence.
```suggestion
        this::class.allSupertypes.first { it.classifier == ViewModelLoader::class }.arguments.first().type?.let { it.classifier as KClass<C> }
```

command-handling/src/main/kotlin/com/ritense/commandhandling/CommandDispatcher.kt
```diff
@@ -0,0 +1,57 @@
...
+            logger.error(ex) { "Unhandled Command error occurred in $commandName - ${ex.message} - ${ex.cause}" }
+            throw ex
+        }
+    }
+
+    fun <T> registerCommandHandler(commandHandler: CommandHandler<Command<T>, T>) {
+        @Suppress("UNCHECKED_CAST")
+        commandHandlers[commandHandler.getCommandType()] =
+            DecoratorBuilder(commandHandler)
+                .decorateWithDefaults()
```
theo-ritense, line 46: Ideally, I'd like to see these decorators provided to this bean, so we can customize the default decorators applied to every registered handler. But we can also add that later.

command-handling/src/test/kotlin/com/ritense/commandhandling/decorator/ExecutionTimeDecoratorTest.kt
```diff
@@ -0,0 +1,27 @@
...
+internal class ExecutionTimeDecoratorTest : BaseTest() {
+
+    @Test
+    fun `should log execution time`(output: CapturedOutput) {
+        val commandHandler = ExecutionTimeDecorator(
+            commandHandler = LambdaCommandHandler()
+        )
+
+        commandHandler.execute(LambdaCommand {
+            Thread.sleep(1000)
```
theo-ritense:
```suggestion
            Thread.sleep(100)
```

command-handling/src/test/kotlin/com/ritense/commandhandling/decorator/LogDecoratorTest.kt
```diff
@@ -0,0 +1,28 @@
...
+internal class LogDecoratorTest : BaseTest() {
+
+    @Test
+    fun `should log command`(output: CapturedOutput) {
+        val commandHandler = LogDecorator(
+            commandHandler = LambdaCommandHandler()
+        )
+
+        commandHandler.execute(LambdaCommand {
+            Thread.sleep(1000)
```
theo-ritense:
```suggestion
            Thread.sleep(100)
```

dependencies/valtimo-dependencies/build.gradle
```diff
@@ -29,6 +29,7 @@ dependencies {
     api(project(":form"))
     api(project(":form-flow"))
     api(project(":form-flow-valtimo"))
+    api(project(":form-view-model"))
```
theo-ritense, line 32: Please add the `command-handling` module as well.

command-handling/gradle/publishing.gradle
```diff
@@ -0,0 +1,35 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+pluginManager.withPlugin('maven-publish') {
+    publishing {
+        publications {
+            maven(MavenPublication) {
+                pom {
+                    name = 'CommandHandling module'
```
theo-ritense:
```suggestion
                    name = 'Command handling module'
```

form-view-model/src/main/kotlin/com/ritense/formviewmodel/validation/OnStartUpViewModelValidator.kt
laurens-ritense: Maybe but please don't

form-view-model/src/main/kotlin/com/ritense/formviewmodel/validation/DataClassPropertiesExtractor.kt
```diff
@@ -0,0 +1,43 @@
...
+
+        // Iterate over each member property of the class
+        for (prop in kClass.memberProperties) {
+            prop.isAccessible = true  // Make private properties accessible
+
+            // Determine the return type of the property
+            val returnType = prop.returnType.classifier as? KClass<*>
+            if (returnType != null && returnType.isData) {
+                // If the return type is a data class, recurse into it
+                results.addAll(extractProperties(returnType, "$prefix${prop.name}."))
```
theo-ritense, line 35: If I read this correctly, having nested data class will result in a key like `prefix-level1.prefix-level2` if the prefix was `prefix-`. I would expect it to be `prefix-level1.level2` instead.

command-handling/src/main/kotlin/com/ritense/commandhandling/CommandDispatcher.kt
laurens-ritense: Let's do that later

form-view-model/src/main/kotlin/com/ritense/formviewmodel/validation/DataClassPropertiesExtractor.kt
laurens-ritense: It's like this 

```
data class Address(
    val street: String,
    val houseNumber: String,
    val zipCode: String
)

data class Person(
    val firstName: String,
    val lastName: String,
    val fullName: String = "$firstName $lastName",
    val age: Int,
    var personalAddress: Address? = null
) : ViewModel {
    override fun update(submission: T, task: CamundaTask?, businessKey: String): ViewModel {
        // code to update the ViewModel
        return this
    }
}
```
becomes:
- firstName
- lastName
- fullName
- age
- personalAddress.street
- personalAddress.houseNumber
- personalAddress.zipCode

theo-ritense (dismissed):

form-view-model/src/main/kotlin/com/ritense/formviewmodel/submission/FormViewModelStartFormSubmissionHandler.kt
```diff
@@ -0,0 +1,75 @@
...
+     * Retrieves the type of the submission.
+     *
+     * This method uses reflection to determine the type of the submission.
+     *
+     * @return the class type of the submission
+     * @throws IllegalArgumentException if the submission type cannot be resolved
+     */
+    @Suppress("UNCHECKED_CAST")
+    fun getSubmissionType(): KClass<T> =
+        this::class.allSupertypes.first().arguments.first().type?.let { it.classifier as KClass<T> }
```
theo-ritense: This does not work when this a class implements it like this, since it is not the first then (see below). The suggestion I made and pushed earlier should work.
```
class Example(): Serializable, FormViewModelStartFormSubmissionHandler<SomeSubmission> { ... }
```

form-view-model/src/main/kotlin/com/ritense/formviewmodel/submission/FormViewModelUserTaskSubmissionHandler.kt
```diff
@@ -0,0 +1,73 @@
...
+     * Retrieves the type of the submission.
+     *
+     * This method uses reflection to determine the type of the submission.
+     *
+     * @return the class type of the submission
+     * @throws IllegalArgumentException if the submission type cannot be resolved
+     */
+    @Suppress("UNCHECKED_CAST")
+    fun getSubmissionType(): KClass<T> =
+        this::class.allSupertypes.first().arguments.first().type?.let { it.classifier as KClass<T> }
```
theo-ritense: This does not work when this a class implements it like this, since it is not the first then (see below). The suggestion I made and pushed earlier should work.
```
class Example(): Serializable, FormViewModelUserTaskSubmissionHandler<SomeSubmission> { ... }
```

form-view-model/src/main/kotlin/com/ritense/formviewmodel/viewmodel/ViewModelLoader.kt
```diff
@@ -0,0 +1,38 @@
...
+@Transactional
+interface ViewModelLoader<T : ViewModel> {
+
+    fun load(task: CamundaTask? = null): T
+
+    fun supports(formName: String) = getFormName() == formName
+
+    @Suppress("UNCHECKED_CAST")
+    fun getViewModelType(): KClass<T> =
+        this::class.allSupertypes.first().arguments.first().type?.let { it.classifier as KClass<T> }
```
theo-ritense: This does not work when this a class implements it like this, since it is not the first then (see below). The suggestion I made and pushed earlier should work.
```
class Example(): Serializable, ViewModelLoader<SomeViewModel> { ... }
```

theo-ritense (approved):

PR #1315, opened 2024-04-18
[BE] Introduce property on FormProcessLink for ViewModel enablement

Tom-Ritense:
Expand existing type with additional property called "ViewModel enabled"
Ensure is can be stored via API and in DB, look the create formlink flow
Ensure existing processlink API exposes this property in the JSON so FE can consume it.


laurens-ritense (changes requested):
give viewModelEnabled a default value of false

laurens-ritense (approved):

PR #1323, opened 2024-04-25
#109886 - [BE] Unable to add `task:` path columns

IvarKoreman-Ritense:
https://ritense.tpondemand.com/entity/109886-be-unable-to-add-task-path

### DIFF
--- a/case/src/main/kotlin/com/ritense/case/service/validations/TaskListColumnValidationUtils.kt
+++ b/case/src/main/kotlin/com/ritense/case/service/validations/TaskListColumnValidationUtils.kt
@@ -27,6 +27,8 @@ import com.ritense.document.service.DocumentDefinitionService
 import com.ritense.valueresolver.ValueResolverService
 import org.zalando.problem.Status
 
+private const val TASK_PREFIX = "task:"
+
 open class TaskListColumnValidationUtils(
     open val taskListColumnRepository: TaskListColumnRepository,
     open val documentDefinitionService: DocumentDefinitionService,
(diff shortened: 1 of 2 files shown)


case/src/main/kotlin/com/ritense/case/service/validations/TaskListColumnValidationUtils.kt
```diff
@@ -83,10 +83,19 @@ open class TaskListColumnValidationUtils(
...
     internal fun isPropertyPathValid(caseDefinitionName: String, taskListColumnDto: TaskListColumnDto) {
-        try {
-            valueResolverService.validateValues(caseDefinitionName, listOf(taskListColumnDto.path))
-        } catch (ex: Exception) {
-            throw InvalidListColumnException(ex.message, Status.BAD_REQUEST)
+        val path = taskListColumnDto.path
+
+        if (path.startsWith("task:", ignoreCase = true)) {
+            val pathWithoutPrefix = path.substring("task:".length)
+            if (!listOf("createTime", "name", "assignee", "dueDate").contains(pathWithoutPrefix)) {
```
Klaas-Ritense, line 92: Could also use the, `CaseTaskProperties.CREATE_TIME` here

case/src/main/kotlin/com/ritense/case/service/validations/TaskListColumnValidationUtils.kt
```diff
@@ -83,10 +83,19 @@ open class TaskListColumnValidationUtils(
 
     @Throws(InvalidListColumnException::class)
     internal fun isPropertyPathValid(caseDefinitionName: String, taskListColumnDto: TaskListColumnDto) {
-        try {
-            valueResolverService.validateValues(caseDefinitionName, listOf(taskListColumnDto.path))
-        } catch (ex: Exception) {
-            throw InvalidListColumnException(ex.message, Status.BAD_REQUEST)
+        val path = taskListColumnDto.path
+
+        if (path.startsWith("task:", ignoreCase = true)) {
```
Klaas-Ritense: Maybe create a prefix for `task:`?

Klaas-Ritense (dismissed):

case/src/main/kotlin/com/ritense/case/service/validations/TaskListColumnValidationUtils.kt
IvarKoreman-Ritense: One of the reasons I refactored the [CaseTaskListSearchService](https://github.com/valtimo-platform/valtimo-backend-libraries/pull/1323/files#diff-ca0c35c8cc513868791e7cebe2b4ce907a79c9867432aa5ee55c2c3324800eaa), was so that I could call `CaseTaskProperties.getNames()`, so I didn't have to replicate the string-literals, and it would also be updated if we add other options to the list. Unfortunately, the `CaseTaskProperties` is part of the process-document module and this `TaskListColumnValidationUtils` is part of the `case` module, so we don't have access to this enum.

Klaas-Ritense (approved):

PR #1333, opened 2024-05-03
Dependency upgrade fix: Downgraded to Liquibase 4.24.0

theo-ritense:
- Changed to `org.hibernate.dialect.PostgreSQLDialect` since `org.hibernate.dialect.PostgreSQL10Dialect` is deprecated
- downgraded to Liquibase 4.24.0 since that is the Spring Boot default

### DIFF
--- a/outbox/outbox-rabbitmq/src/test/resources/config/application-postgresql.yml
+++ b/outbox/outbox-rabbitmq/src/test/resources/config/application-postgresql.yml
@@ -7,7 +7,7 @@ spring:
         hikari:
             auto-commit: false
     jpa:
-        database-platform: org.hibernate.dialect.PostgreSQL10Dialect
+        database-platform: org.hibernate.dialect.PostgreSQLDialect
         database: postgresql
 
 valtimo:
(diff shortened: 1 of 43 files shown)


ivo-ritense (approved):

outbox/outbox-rabbitmq/src/test/resources/config/application-postgresql.yml
```diff
@@ -7,7 +7,7 @@ spring:
         hikari:
             auto-commit: false
     jpa:
-        database-platform: org.hibernate.dialect.PostgreSQL10Dialect
```
Klaas-Ritense, line 10: I remember adding the PostgreSQL10Dialect to support the implementation specific query that is used in `PostgresOutboxMessageRepository`. I'm not sure how the outbox tests are succeeding with this change

theo-ritense: So, when you take a look at the `PostgreSQL10Dialect`, you'll see the following:
```
public class PostgreSQL10Dialect extends PostgreSQLDialect {
    public PostgreSQL10Dialect() {
        super(DatabaseVersion.make(10));
    }
}
```

This is the constructor of `PostgreSQLDialect`"
```
public PostgreSQLDialect() {
    this(MINIMUM_VERSION); //MINIMUM_VERSION = DatabaseVersion.make(10);
}
```
That makes them equal in my book.

PR #1345, opened 2024-05-14
109484: CW admin add/edit tab

theo-ritense:


case/src/main/kotlin/com/ritense/case/security/config/CaseHttpSecurityConfigurer.kt
```diff
@@ -70,5 +71,6 @@ class CaseHttpSecurityConfigurer : HttpSecurityConfigurer {
         private const val MANAGEMENT_CASE_LIST_COLUMN_URL = "/api/management/v1/case/{caseDefinitionName}/list-column"
         private const val MANAGEMENT_TASK_LIST_COLUMN_URL = "/api/management/v1/case/{caseDefinitionName}/task-list-column"
         private const val MANAGEMENT_TAB_URL = "/api/management/v1/case-definition/{caseDefinitionName}/tab"
+        private const val MANAGEMENT_WIDGET_TAB_URL = "/api/management/v1/case-definition/{caseDefinitionName}/widget-tab"
```
marijnritense, line 74: I know we do this in more places, but when you use versioning this gets a bit iffy. Could you either rename this variable to `MANAGEMENT_WIDGET_TAB_URL_v1` or split this up into smaller parts, so the version still has to be specified?

case/src/main/kotlin/com/ritense/case/service/CaseTabService.kt
```diff
@@ -70,9 +73,9 @@ class CaseTabService(
             ?: throw NoSuchElementException("Case definition with name $caseDefinitionName does not exist!")
 
         val currentTabs = getCaseTabs(caseDefinitionName)
-        val tabWithKeyExists = currentTabs.filter { tab ->
-            tab.id.key.equals(caseTabDto.key)
-        }.isNotEmpty()
+        val tabWithKeyExists = currentTabs.any { tab ->
```
marijnritense, line 76: Wouldn't you just want to add a method to currentTabs instead of do it here? Getting a certain tab seems like a useful function

theo-ritense: This check is pretty specific for when a new tab is created. 
Also, the changes I made were merely a fix to the incorrect usage of `.equals()` in Kotlin and suggested style improvements by the IDE.

case/src/main/kotlin/com/ritense/case/security/config/CaseHttpSecurityConfigurer.kt
theo-ritense: As discussed, the other options aren't any more readable than the code above

marijnritense (approved):

PR #1346, opened 2024-05-14
Feature/111002 case widgets admin tab widgets

marijnritense:


case/src/main/kotlin/com/ritense/case_/rest/CaseWidgetTabManagementResource.kt
```diff
@@ -41,4 +44,16 @@ class CaseWidgetTabManagementResource(
         val widgetTab = caseWidgetTabService.getWidgetTab(caseDefinitionName, tabKey)
         return ResponseEntity.ofNullable(widgetTab)
     }
+
+    @PostMapping("/v1/case-definition/{caseDefinitionName}/widget-tab/{tabKey}")
+    fun getCaseTabWidget(
```
theo-ritense: This method name does not match the action (update)

case/src/main/kotlin/com/ritense/case_/service/CaseWidgetTabService.kt
```diff
@@ -43,4 +50,37 @@ class CaseWidgetTabService(
         return caseWidgetTabRepository.findByIdOrNull(CaseTabId(caseDefinitionName, key))
             ?.let { CaseWidgetTabDto.of(it) }
     }
+
+    fun updateWidgetTab(tabDto: CaseWidgetTabDto): CaseWidgetTabDto? {
+        denyAuthorization()
+
+        val caseWidgetTab = caseWidgetTabRepository.findById(CaseTabId(tabDto.caseDefinitionName, tabDto.key))
```
theo-ritense: Could've used `findByIdOrNull` here to keep it more Kotlin-style, but this works too.

case/src/test/kotlin/com/ritense/case_/service/CaseWidgetTabServiceIntTest.kt
```diff
@@ -66,4 +69,150 @@ class CaseWidgetTabServiceIntTest @Autowired constructor(
...
+                    tabKey,
+                    widgets = listOf(
+                        CaseWidgetTabWidgetDto("widget-1", "Widget 1", 0, false),
+                        CaseWidgetTabWidgetDto("widget-2", "Widget 2", 1, true)
+                    )
+                )
+            )
+        }
+
+        val optionalWidgetTab = caseWidgetTabRepository.findById(CaseTabId(caseDefinitionName, tabKey))
```
theo-ritense: This would be way nicer if you used `findByIdOrNull` instead. Same for all other usages of `.findById` in this file.

case/src/main/kotlin/com/ritense/case_/domain/tab/CaseWidgetTabWidget.kt
```diff
@@ -0,0 +1,51 @@
...
+import jakarta.persistence.Table
+
+@Entity
+@Table(name = "case_widget_tab_widget")
+@Inheritance(strategy = InheritanceType.SINGLE_TABLE)
+@DiscriminatorColumn(
+    name = "case_widget_type",
+    discriminatorType = DiscriminatorType.STRING
+)
+data class CaseWidgetTabWidget(
```
theo-ritense: How can we use the discriminator type when this is a (closed) data class? I don't see any test that proves that the inheritance is working.

theo-ritense: Fixed it.

theo-ritense (approved):

PR #1347, opened 2024-05-14
Fix/form view model validation

laurens-ritense:


form-view-model/src/main/kotlin/com/ritense/formviewmodel/validation/OnStartUpViewModelValidator.kt
```diff
@@ -99,6 +103,25 @@ class OnStartUpViewModelValidator(
         return results
     }
 
+    // This transforms [a.b.c, a.b.d, a.e] into [a, b, c, d, e]
+    private fun transformListOfProperties(originalList: List<String>): List<String> {
```
Tom-Ritense: Why? what does this make better? 
[address.number] is a really clear notation i find.

laurens-ritense: It does not match what formDefinition.inputFields returns this does

Tom-Ritense: Can you give some examples? lets take this offline

Tom-Ritense (approved):

PR #1353, opened 2024-05-21
Added case widget user tab

theo-ritense:


case/src/main/kotlin/com/ritense/case_/service/CaseWidgetTabService.kt
```diff
@@ -69,6 +78,34 @@ class CaseWidgetTabService(
...
+
+        //TODO: Do some authorization check on the widget here?
+
+        return caseWidgetDataProviders.firstNotNullOf { provider ->
+            provider.getData(widgetTab, widgetConfig)
+        }
+    }
+
+    private fun checkCaseTabAccess(caseDefinitionName: String, key: String, action: Action<CaseTab>) {
+        if (!AuthorizationContext.ignoreAuthorization) {
```
marijnritense: The ignore authorization is done on another level (by PBAC) already. Why is this check needed here?

theo-ritense: Performance reasons. We're only using the result of `caseTabRepository.findByIdOrNull` for the PBAC check.

marijnritense (dismissed):

marijnritense (approved):

PR #1365, opened 2024-05-24
Case/109496 case widget table

theo-ritense:


case/src/main/kotlin/com/ritense/case_/widget/table/TableCaseWidgetDataProvider.kt
```diff
@@ -0,0 +1,86 @@
...
+                    }
+
+                    column.key to value
+                }
+            }
+
+        return PageImpl(result, pageable, collection.size().toLong())
+    }
+
+    class InvalidCollectionException : RuntimeException("Invalid collection value!")
```
marijnritense: You'll reuse this with the collection case widget as well I think. Could you move this to a separate file?

PR #1370, opened 2024-05-27
#109359 Limit input accepted by form-flow

Klaas-Ritense:


form-flow-valtimo/src/main/kotlin/com/ritense/valtimo/formflow/service/FormFlowValtimoService.kt
```diff
@@ -0,0 +1,65 @@
...
+        val currentStepTypeProperties = formFlowInstance.getCurrentStep().definition.type.properties
+        if (currentStepTypeProperties !is FormStepTypeProperties) {
+            return submissionData
+        }
+
+        val jsonPatchBuilder = JsonPatchBuilder()
+        val verifiedSubmissionData = objectMapper.createObjectNode()
+
+        val validJsonPointers = formDefinitionService.getFormDefinitionByName(currentStepTypeProperties.definition)
+            .orElseThrow().inputFields
```
marijnritense, line 50: Does this also work with nested properties?

marijnritense:
I think this PR needs to go to `next-major` as it's a breaking change.

form-flow-valtimo/src/main/kotlin/com/ritense/valtimo/formflow/service/FormFlowValtimoService.kt
theo-ritense: @marijnritense Not sure if I understand the question correctly, but the `inputFields` method also returns fields within (sub)containers.

theo-ritense (approved):

PR #1371, opened 2024-05-27
Added widget authorization

ivo-ritense:


authorization/src/main/kotlin/com/ritense/authorization/permission/condition/ReflectingPermissionCondition.kt
```diff
@@ -16,12 +16,25 @@
...
     protected fun findEntityFieldValue(entity: Any, field: String): Any? {
         var currentEntity: Any? = entity
         val fields = field.split('.')
         fields.forEachIndexed { index, value ->
-            val declaredField = currentEntity!!.javaClass.getDeclaredField(value)
+            var classToSearch: Class<*>? = entity.javaClass
+            var declaredField: Field? = null
+
+            while (declaredField == null && classToSearch != null) {
+                declaredField = classToSearch.declaredFields.firstOrNull { it.name == value }
```
theo-ritense, line 33: Maybe not part of this story perse since it was already there, but why are we checking declared fields instead of just the public fields? This means we can never change a private field without it effectively being a breaking change.

authorization/src/main/kotlin/com/ritense/authorization/permission/condition/ReflectingPermissionCondition.kt
```diff
@@ -16,12 +16,25 @@
...
+                if (declaredField == null)
+                    classToSearch = classToSearch.superclass
+            }
+
+            if (declaredField == null) {
+                throw IllegalArgumentException("Field $field not found in class ${entity.javaClass}")
+            }
             declaredField.trySetAccessible()
```
theo-ritense, line 41: Usage of trySetAccessible is discouraged, and would not be needed if we only supported public fields (see comment above)

case/src/main/kotlin/com/ritense/case/repository/CaseTabDocumentDefinitionMapper.kt
```diff
@@ -0,0 +1,53 @@
...
+import com.ritense.case.domain.CaseTabId
+import com.ritense.document.domain.impl.JsonSchemaDocumentDefinition
+import com.ritense.document.domain.impl.JsonSchemaDocumentDefinitionId
+import com.ritense.document.service.DocumentDefinitionService
+import jakarta.persistence.criteria.AbstractQuery
+import jakarta.persistence.criteria.CriteriaBuilder
+import jakarta.persistence.criteria.Root
+
+class CaseTabDocumentDefinitionMapper(
+    private val documentDefinitionService: DocumentDefinitionService
```
theo-ritense, line 32: I believe we have as story on the usage of services within mappers. This will generate a ton of audit messages just for checking permissions.

case/src/main/kotlin/com/ritense/case_/rest/dto/CaseWidgetTabDto.kt
```diff
@@ -29,11 +29,16 @@ data class CaseWidgetTabDto(
...
+        fun of(
+            tab: CaseWidgetTab,
+            widgetMappers: List<CaseWidgetMapper<CaseWidgetTabWidget, CaseWidgetTabWidgetDto>>,
+            permissionCheck: (CaseWidgetTabWidget) -> Boolean
+        ): CaseWidgetTabDto {
             return CaseWidgetTabDto(
                 tab.id.caseDefinitionName,
                 tab.id.key,
                 widgets = tab.widgets
+                    .filter { permissionCheck(it) }
```
theo-ritense, line 41: Alternatively, we could do the check in the widgetMappers and return null from there if the check fails. Not sure if it's better than this though.

authorization/src/main/kotlin/com/ritense/authorization/permission/condition/ReflectingPermissionCondition.kt
marijnritense: It's needed as might be applying AC to classes we do not create ourselves.

case/src/main/kotlin/com/ritense/case_/repository/CaseWidgetTabWidgetSpecification.kt
```diff
@@ -0,0 +1,60 @@
...
+                    criteriaBuilder,
+                    authRequest.resourceType,
+                    queryDialectHelper
+                )
+            }.toList()
+        return combinePredicates(criteriaBuilder, predicates)
+    }
+
+    override fun identifierToEntity(identifier: String): CaseWidgetTabWidget {
+        TODO("Not yet implemented")
```
marijnritense, line 57: @theo-ritense Should this be done for the feature, or do we just pick this up sometime in the future when someone asks for this?

theo-ritense: If we can think of a use case, it might be better to implement it now.

authorization/src/main/kotlin/com/ritense/authorization/permission/condition/ReflectingPermissionCondition.kt
theo-ritense: Again: if we only supported public fields this would not be needed right? I don't see why we would need to support non-public fields.

theo-ritense (dismissed):

Klaas-Ritense (dismissed):

Klaas-Ritense (approved):

PR #1374, opened 2024-05-28
Enable viewmodel by order of ProcessLinkActivityHandler

sirolf2009:


sirolf2009:
https://github.com/valtimo-platform/valtimo-backend-libraries/pull/1377

PR #1401, opened 2024-06-06
Major 12 fixes

Klaas-Ritense:
- Fix liquibase checksum
- Also disable fast-check-cashing. See [this PR](https://github.com/valtimo-platform/valtimo-backend-libraries/pull/1389) for more info
- Make notificationService a processBean
- Add CamundaBeansPlugin to disable authorization when whitelisting is false
- Add `@Primary` on 2nd repositoryService to prevent 'conflicting dependency' error
- Add Camunda migration scripts


theo-ritense (dismissed):

theo-ritense (dismissed):

theo-ritense (approved):

core/src/main/kotlin/com/ritense/valtimo/CamundaWhitelistedBeansPlugin.kt
```diff
@@ -0,0 +1,43 @@
...
+
+package com.ritense.valtimo
+
+import mu.KotlinLogging
+import org.camunda.bpm.engine.impl.cfg.AbstractProcessEnginePlugin
+import org.camunda.bpm.engine.impl.cfg.ProcessEngineConfigurationImpl
+import org.camunda.bpm.engine.spring.SpringExpressionManager
+import org.springframework.context.ApplicationContext
+
+class CamundaWhitelistedBeansPlugin(
```
marijnritense, line 25: What was the reason behind splitting this off? To keep more focus to the plugin?

Klaas-Ritense: Its to fix a bug. By default, Valtimo doesn't do authorization on service tasks because those are always executed by the system. But the bug is; when you have `bean-whitelisting` set to false, suddenly this doesn't work anymore and you get authorization errors on service-tasks.

marijnritense: Makes sense, thanks for the explanation!

PR #1414, opened 2024-06-10
Story/109503 fixes

theo-ritense:


case/src/main/kotlin/com/ritense/case/repository/CaseTabDocumentDefinitionMapper.kt
```diff
@@ -27,12 +27,13 @@ import com.ritense.document.service.DocumentDefinitionService
...
 import jakarta.persistence.criteria.CriteriaBuilder
 import jakarta.persistence.criteria.Root
+import kotlin.jvm.optionals.getOrNull
 
 class CaseTabDocumentDefinitionMapper(
     private val documentDefinitionService: DocumentDefinitionService
 ) : AuthorizationEntityMapper<CaseTab, JsonSchemaDocumentDefinition> {
     override fun mapRelated(entity: CaseTab): List<JsonSchemaDocumentDefinition> {
-        return runWithoutAuthorization { listOf(documentDefinitionService.findLatestByName(entity.id.caseDefinitionName) as JsonSchemaDocumentDefinition) }
+        return runWithoutAuthorization { listOfNotNull(documentDefinitionService.findLatestByName(entity.id.caseDefinitionName).map { it as JsonSchemaDocumentDefinition}.getOrNull()) }
```
Klaas-Ritense: Shouldn't we throw an error when a documentDefinition doesn't exist?

theo-ritense: I'm not sure. The contract expects a List as a result, and a List can be empty. It is up to the consumer to handle an empty list imho. @marijnritense ?

marijnritense: If a provided case definition name does not lead to any results, I would say there's something weird going on, and throwing an exception makes sense.

Klaas-Ritense (approved):

PR #1427, opened 2024-06-12
Story/113212 cw authorization

marijnritense:


authorization/src/main/kotlin/com/ritense/authorization/permission/Permission.kt
```diff
@@ -57,27 +57,82 @@ data class Permission(
...
+        queryDialectHelper: QueryDialectHelper,
+    ): Predicate = toPredicate(root, query, criteriaBuilder, resourceType, queryDialectHelper, null, null)
+
     fun <T : Any> toPredicate(
         root: Root<T>,
         query: AbstractQuery<*>,
         criteriaBuilder: CriteriaBuilder,
         resourceType: Class<T>,
-        queryDialectHelper: QueryDialectHelper
+        queryDialectHelper: QueryDialectHelper,
```
theo-ritense, line 128: My ocd doesn't like this `queryDialectHelper` in the middle, and since you've created a secondary method anyway to make it non-breaking, you might as well reorder this one.

authorization/src/main/kotlin/com/ritense/authorization/permission/Permission.kt
```diff
@@ -92,4 +147,16 @@ data class Permission(
...
+    private fun <U> appliesInContext(
+        contextResourceType: Class<U>?,
+        contextEntity: Any?
+    ): Boolean {
+        return (this.contextResourceType == null
+            || (contextResourceType == this.contextResourceType
+                && (contextConditionContainer?.let { container ->
+                    container.conditions
+                        .all { it.isValid(contextEntity!!) }
+                    } == true)))
```
theo-ritense, line 160: The `== true` isn't needed, is it?

case/src/main/kotlin/com/ritense/case_/service/CaseWidgetTabService.kt
```diff
@@ -69,6 +71,27 @@ class CaseWidgetTabService(
 
     }
 
+    @Transactional(readOnly = true)
```
theo-ritense: Doesn't this `readOnly` break our outbox? If you didn't already, please test this without a parent transaction (like we do in tests).

case/src/main/kotlin/com/ritense/case_/rest/CaseWidgetTabResource.kt
```diff
@@ -36,12 +37,12 @@ class CaseWidgetTabResource(
     private val caseWidgetTabService: CaseWidgetTabService
 ) {
 
-    @GetMapping("/v1/case-definition/{caseDefinitionName}/widget-tab/{tabKey}")
+    @GetMapping("/v1/document/{documentId}/widget-tab/{tabKey}")
```
theo-ritense, line 40: I'm not seeing a change to the web security config. Did I miss it?

theo-ritense (approved):

PR #1431, opened 2024-06-17
feat: show tasks on case tabs

mbritense:
closes https://ritense.tpondemand.com/entity/115539-be-tasks-on-case-tabs

### DIFF
--- a/case/src/main/kotlin/com/ritense/case/domain/CaseTab.kt
+++ b/case/src/main/kotlin/com/ritense/case/domain/CaseTab.kt
@@ -42,6 +42,7 @@ data class CaseTab(
     val createdOn: LocalDateTime? = LocalDateTime.now(),
 
     val createdBy: String? = null,
+    val showTasks: Boolean = false
 ) {
     constructor(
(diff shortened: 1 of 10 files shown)


case/src/main/kotlin/com/ritense/case/domain/CaseTab.kt
```diff
@@ -42,14 +42,16 @@ data class CaseTab(
...
 ) {
     constructor(
         id: CaseTabId,
         name: String?,
         tabOrder: Int,
         type: CaseTabType,
         contentKey: String,
-    ) : this(id, name, tabOrder, type, contentKey, LocalDateTime.now(), null)
+        showTasks: Boolean
+    ) : this(id, name, tabOrder, type, contentKey, LocalDateTime.now(), null, showTasks)
```
Klaas-Ritense:
```suggestion
    ) : this(id, name, tabOrder, type, contentKey, LocalDateTime.now(), null, false)
```
To avoid a breaking change

Klaas-Ritense (approved):

PR #1469, opened 2024-07-12
117355 : Fix empty mail recipients list

theo-ritense:


mail/src/main/kotlin/com/ritense/mail/domain/filters/BlacklistFilter.kt
```diff
@@ -41,17 +42,23 @@ class BlacklistFilter(
...
+    private fun <T: HasRecipients> doFilterInternal(mailMessage: T): Optional<T> {
+        mailMessage
             .recipients
             .filterBy { !blacklistService.isBlacklisted(it.email.get()) }
-        return Optional.of(templatedMailMessage)
+
+        return if (mailMessage.recipients.isPresent) {
+            Optional.of(mailMessage)
+        } else {
+            Optional.empty()
```
ivo-ritense, line 63: Do we want to add debug logging so users might understand why their mail is not being sent?

theo-ritense: fixed

ivo-ritense (approved):

PR #1471, opened 2024-07-15
Feature/user identifier toggle

marijnritense:


authorization/src/main/kotlin/com/ritense/authorization/permission/condition/PermissionConditionValueResolver.kt
```diff
@@ -26,6 +26,7 @@ object PermissionConditionValueResolver {
                 "\${currentUserId}" -> UserManagementServiceHolder.currentInstance.currentUser.id
                 "\${currentUserEmail}" -> UserManagementServiceHolder.currentInstance.currentUser.email
                 "\${currentUserRoles}" -> UserManagementServiceHolder.currentInstance.currentUser.roles
+                "\${currentUserIdentifier}" -> UserManagementServiceHolder.currentInstance.currentUser.userIdentifier
```
theo-ritense, line 29: Each call to `currentUser` will trigger some expensive calls to Keycloak. Please 'cache' it locally (Take care not to call it always, because it might not be needed).

Edit: I was wrong. This method handles only one value at a time anyways.

contract/src/main/kotlin/com/ritense/valtimo/contract/OauthHolder.kt
```diff
@@ -0,0 +1,33 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.ritense.valtimo.contract
+
+import com.ritense.valtimo.contract.config.ValtimoProperties.Oauth
+
+class OauthHolder(oauth: Oauth) {
```
theo-ritense: Maybe rename this to `OauthPropertiesHolder` or `OauthConfigHolder`? It's not holding the JWT itself, so the current name is mildly confusing.

contract/src/main/kotlin/com/ritense/valtimo/contract/OauthHolder.kt
```diff
@@ -0,0 +1,33 @@
...
+
+    init {
+        Companion.oauth = oauth
+    }
+
+    companion object {
+        private var oauth: Oauth? = null
+
+        val currentInstance: Oauth
```
theo-ritense: Please add `@JvmField` here to avoid the `.Companion` notation in Java

theo-ritense (approved):

PR #1480, opened 2024-07-22
Added support for task: prefix on task filter

theo-ritense:
- Added support for task: prefix on task filter
- Added zgw module to dependency-versions and gzac-dependencies


process-document/src/main/kotlin/com/ritense/processdocument/service/CaseTaskListSearchService.kt
```diff
@@ -59,6 +59,12 @@ import jakarta.persistence.criteria.Order
 import jakarta.persistence.criteria.Path
 import jakarta.persistence.criteria.Predicate
 import jakarta.persistence.criteria.Root
+import org.apache.commons.lang3.NotImplementedException
```
marijnritense, line 62: Why the shuffle of imports?

marijnritense (approved):

PR #1485, opened 2024-07-29
Improve error message: valtimo.plugin.encryption-secret

Klaas-Ritense:


plugin/src/main/kotlin/com/ritense/plugin/configuration/PluginAutoConfiguration.kt
```diff
@@ -177,6 +177,11 @@ class PluginAutoConfiguration {
         @Value("\${valtimo.plugin.encryption-secret}")
         secret: String
     ): EncryptionService {
+        val secretByteLength = secret.toByteArray().size
+        require(secretByteLength == 16 || secretByteLength == 24 || secretByteLength == 32) {
+            "Invalid byte length '$secretByteLength'. Expected property 'valtimo.plugin.encryption-secret' or 'VALTIMO_PLUGIN_ENCRYPTION-SECRET' to have a byte length of 32"
```
theo-ritense: The message does not reflect the checks on 16 and 24 bytes.

Klaas-Ritense:
```suggestion
            "Invalid AES key length of '$secretByteLength' bytes. Expected property 'valtimo.plugin.encryption-secret' or 'VALTIMO_PLUGIN_ENCRYPTION-SECRET' to have a length of 16 or 24 or 32 bytes"
```

marijnritense (approved):

PR #1498, opened 2024-08-08
#115546 OAuth 2 for Keycloak

Klaas-Ritense:
Old configuration still works:
`keycloak.auth-server-url: http://localhost:8081/auth`

But when the new configuration is used, the application will switch to use the new OAuth 2 authentication:
`spring.security.oauth2.client.provider.keycloakjwt.issuer-uri: http://localhost:8081/auth/realms/valtimo`


contract/src/main/kotlin/com/ritense/valtimo/contract/security/config/oauth2/NoOAuth2ClientsConfiguredCondition.kt
```diff
@@ -0,0 +1,54 @@
...
+import org.springframework.boot.autoconfigure.condition.ConditionOutcome
+import org.springframework.boot.autoconfigure.condition.SpringBootCondition
+import org.springframework.boot.autoconfigure.security.oauth2.client.OAuth2ClientProperties
+import org.springframework.boot.context.properties.bind.Bindable
+import org.springframework.boot.context.properties.bind.Binder
+import org.springframework.context.annotation.ConditionContext
+import org.springframework.core.env.Environment
+import org.springframework.core.type.AnnotatedTypeMetadata
+
+// The opposite of: org.springframework.boot.autoconfigure.security.oauth2.client.ClientsConfiguredCondition
```
theo-ritense, line 28: That class states `Condition that matches if any spring.security.oauth2.client.registration properties are defined.`

It does not explain how it works, of why we need the negated version of it.

core/src/main/java/com/ritense/valtimo/autoconfigure/AuthenticationAutoConfiguration.java
```diff
@@ -21,14 +21,17 @@
...
 import org.springframework.boot.autoconfigure.AutoConfiguration;
 import org.springframework.boot.autoconfigure.condition.ConditionalOnMissingBean;
 import org.springframework.context.annotation.Bean;
+import org.springframework.context.annotation.Conditional;
 
 @AutoConfiguration
 public class AuthenticationAutoConfiguration {
 
     @Bean
+    @Conditional(NoOAuth2ClientsConfiguredCondition.class)
```
theo-ritense, line 34: Does the application still start/work then this bean is being skipped?

theo-ritense (commented):
I see you have tried to make it non-breaking. Did you test if it really is?

core/src/main/java/com/ritense/valtimo/autoconfigure/AuthenticationAutoConfiguration.java
Klaas-Ritense: Yes. I tested it thoroughly on my local machine.

Basically, the tokenAuthenticationService together with some other Valtimo services were needed to handle the authentication. This can now all be replaced by OAuth2. When an application is now configured for OAuth2, it won't need those Valtimo services anymore.

contract/src/main/kotlin/com/ritense/valtimo/contract/security/config/oauth2/NoOAuth2ClientsConfiguredCondition.kt
Klaas-Ritense: When an application has configured the `spring.security.oauth2.client.registration` property, Valtimo will switch out its old authentication services/bean to use then new OAuth2 authentication beans.

theo-ritense (approved):

PR #1512, opened 2024-08-20
Fix for total count of filtered task list result

maarten-ritense:
The task list total count query was incorrect due to that a group by condition was added to the query which resulted in a result of X rows with each a count of 1

### Breaking changes

- [X] The contribution only contains changes that are not breaking.

### Documentation

- [X] Not applicable

>

- [X] Not applicable

### Tests
- [X] Not applicable

- [X] Not applicable

### Security
- [X] Not applicable

- [X] Not applicable

- [X] Not applicable

### Dependencies
- [X] Not applicable

### DIFF
--- a/authorization/src/main/kotlin/com/ritense/authorization/utils/QueryUtils.kt
+++ b/authorization/src/main/kotlin/com/ritense/authorization/utils/QueryUtils.kt
@@ -1,3 +1,19 @@
+/*
+ * Copyright 2015-2024 Ritense BV, the Netherlands.
+ *
+ * Licensed under EUPL, Version 1.2 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * https://joinup.ec.europa.eu/collection/eupl/eupl-text-eupl-12
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" basis,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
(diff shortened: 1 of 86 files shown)


authorization/src/main/kotlin/com/ritense/authorization/utils/QueryUtils.kt
Klaas-Ritense, line 1: Missing copyright header

document/src/main/java/com/ritense/document/service/impl/JsonSchemaDocumentSearchService.java
```diff
@@ -190,25 +190,15 @@ private Page<JsonSchemaDocument> search(QueryWhereBuilder queryWhereBuilder, Pag
...
-        List<Long> countResultList = entityManager.createQuery(countQuery).getResultList();
-
-        long count = 0L;
-
-        if (!countResultList.isEmpty()) {
-            count = countResultList.size();
-        }
-
-        return count;
+        return entityManager.createQuery(countQuery).getSingleResult();
```
Klaas-Ritense, line 201: This changes the outcome. But there isn't a single test that breaks. 
I wonder why a count query used `.size();` 

@ivo-ritense do you know?

ivo-ritense: It's because we no longer do a group by for count queries. This means we get a single result instead of a list of results, so we no longer need the `size()`. The `countDistinct` is used to prevent duplicate results caused by the join on other tables. This is closer to the 'proper' SQL query that should be used for something like this.

ivo-ritense (approved):

PR #1522, opened 2024-08-26
POC: Form Flow Breadcrumbs

Klaas-Ritense:


mbritense (dismissed):

ivo-ritense (dismissed):

ivo-ritense (dismissed):

form-flow/src/main/kotlin/com/ritense/formflow/service/FormFlowService.kt
```diff
@@ -100,4 +102,31 @@ class FormFlowService(
...
+        return historicBreadcrumbs + futureBreadcrumbs
+    }
+
+    private fun getFutureSteps(instance: FormFlowInstance): List<FormFlowStep> {
+        return getFutureSteps(instance.getHistory().last().definition)
+    }
+
+    private fun getFutureSteps(step: FormFlowStep, result: MutableList<FormFlowStep> = mutableListOf()): List<FormFlowStep> {
+        step.nextSteps.forEach { nextStep ->
+            val featureStep = step.id.formFlowDefinition!!.steps.first { it.id.key == nextStep.step }
```
marijnritense: So the first step will be the happy path? for every form flow step? I'm not sure if that is explicit enough.

marijnritense: Also, shouldn't this be `futureStep` rather than `featureStep`?

Klaas-Ritense: The first step is indeed the happy path. This is explained in the documentation.

Also: fixed typo

marijnritense (approved):

PR #1525, opened 2024-08-28
#109683: Fix submissionData editing

Klaas-Ritense:


marijnritense (approved):
Approving this, but let's ask Thomas if he's okay with considering this a non-breaking change as soon as he gets back.

PR #1531, opened 2024-08-30
Added logging context annotation

ivo-ritense:


core/src/main/kotlin/com/ritense/valtimo/logging/UserLoggingFilter.kt
```diff
@@ -0,0 +1,52 @@
...
+class UserLoggingFilter(
+    private val userManagementService: UserManagementService
+) : OncePerRequestFilter() {
+
+    override fun doFilterInternal(
+        request: HttpServletRequest,
+        response: HttpServletResponse,
+        filterChain: FilterChain
+    ) {
+        val userIdentifier = userManagementService.getCurrentUser()?.getUserIdentifier()
```
Klaas-Ritense, line 37: For every API call, a request to Keycloak is made. Making it a bit expensive. Is it possible to get the userIdentifier from the JWT token?

ivo-ritense: Currently this is unfortunately not possible to do differently without a breaking change. We should look into improving this soon.

marijnritense (approved):

core/src/main/kotlin/com/ritense/valtimo/logging/UserLoggingFilter.kt
Klaas-Ritense: https://github.com/valtimo-platform/valtimo-backend-libraries/pull/1544

PR #1532, opened 2024-09-02
#118275 storing technical logs

Klaas-Ritense:


logging/src/main/kotlin/com/ritense/logging/domain/LogginEvent.kt
```diff
@@ -0,0 +1,92 @@
...
+import jakarta.persistence.Table
+import org.springframework.data.annotation.Immutable
+import java.time.Instant
+import java.time.LocalDateTime
+import java.time.ZoneId
+
+@Immutable
+@Entity
+@Table(name = "logging_event")
+class LoggingEvent(
```
marijnritense: TItle of the class doesn't seem to match the filename. Could you change the filename? Same goes for the other domain classes.

logging/src/main/resources/config/logging/logback/valtimo-database-appender.xml
```diff
@@ -0,0 +1,50 @@
...
+            </dataSource>
+        </connectionSource>
+        <insertHeaders>true</insertHeaders>
+    </appender>
+
+    <appender name="VALTIMODBASYNC" class="ch.qos.logback.classic.AsyncAppender">
+        <appender-ref ref="VALTIMODB" />
+        <includeCallerData>false</includeCallerData>
+        <maxFlushTime>5000</maxFlushTime>
+        <neverBlock>true</neverBlock>
```
marijnritense, line 48: Leaving this as a note, no need to do anything with it right now.

We should probably discuss if async is the way to go. Performance wise this is probably a great idea, but that means if a log cant be stored, then we don't have a papertrail. Bit of a weird situation. but I think sticking with async is the right call for now.

logging/src/main/resources/config/liquibase/changelog/logback/mysql.sql
```diff
@@ -0,0 +1,53 @@
+# Logback: the reliable, generic, fast and flexible logging framework.
```
ivo-ritense, line 1: How does the license work in this case? Can we just include licensed code in our repo? Alternatively, is this included in the logback artifact and can we reference that?

Klaas-Ritense: I've checked it and its allowed. 

It's not possible to reference the existing files in their library, due to small but necessary changed made to these files.

logging/src/main/kotlin/com/ritense/logging/domain/LogginEvent.kt
Klaas-Ritense: Changed

logging/src/main/resources/config/logging/logback/valtimo-database-appender.xml
Klaas-Ritense: Luckily, this logback file creates 2 appenders: `VALTIMODB` and `VALTIMODBASYNC`. 

But yes, the `VALTIMODBASYNC` performance is much better

ivo-ritense (approved):

PR #1536, opened 2024-09-09
#118278 Periodically cleaning up technical logs

Klaas-Ritense:


logging/src/main/kotlin/com/ritense/logging/LoggingAutoConfiguration.kt
```diff
@@ -56,6 +58,23 @@ class LoggingAutoConfiguration {
         return LoggableResourceAspect()
     }
 
+    @Bean
+    @ConditionalOnMissingBean(LoggingEventDeletionService::class)
+    fun loggingEventDeletionService(
+        @Value("\${valtimo.logging.retentionInMinutes:10080}") retentionInMinutes: Long,
```
ivo-ritense: I think this might be too low for a default value. Maybe we should go for a month?

logging/src/main/kotlin/com/ritense/logging/service/LoggingEventDeletionService.kt
```diff
@@ -0,0 +1,53 @@
...
+    private val loggingEventExceptionRepository: LoggingEventExceptionRepository,
+) {
+
+    @Scheduled(
+        fixedRateString = "\${valtimo.logging.retentionInMinutes:10080}",
+        timeUnit = MINUTES
+    )
+    fun deleteOldLoggingEvents() {
+        val retentionDateTime = LocalDateTime.now() - Duration.ofMinutes(retentionInMinutes)
+        loggingEventPropertyRepository.delete(LoggingEventPropertySpecificationHelper.byOlderThan(retentionDateTime))
```
ivo-ritense, line 49: Doesn't this cascade?

ivo-ritense:
There are no tests in this PR

logging/src/main/kotlin/com/ritense/logging/service/LoggingEventDeletionService.kt
Klaas-Ritense: No

logging/src/main/kotlin/com/ritense/logging/LoggingAutoConfiguration.kt
Klaas-Ritense: Changed

logging/src/main/kotlin/com/ritense/logging/service/LoggingEventDeletionService.kt
ivo-ritense: As discussed: There is probably a reason for this, and we assume it's related to the large amount of inserts this should support and foreign keys might slow it down.

ivo-ritense (dismissed):

ivo-ritense (approved):

PR #1537, opened 2024-09-10
Fix: Notificaties api plugin creation calls Valtimo with anonymous user

Klaas-Ritense:


core/src/main/kotlin/com/ritense/valtimo/logging/UserLoggingFilter.kt
```diff
@@ -34,9 +35,13 @@ class UserLoggingFilter(
...
-        val userIdentifier = userManagementService.getCurrentUser()?.getUserIdentifier()
-        if (userIdentifier != null) {
-            MDC.put(LoggingConstants.MDC_USER_ID_KEY, userIdentifier)
+        try {
+            val userIdentifier = userManagementService.getCurrentUser()?.getUserIdentifier()
+            if (userIdentifier != null) {
+                MDC.put(LoggingConstants.MDC_USER_ID_KEY, userIdentifier)
+            }
+        } catch (e: Exception) {
+            logger.error("Failed to add logging context '${LoggingConstants.MDC_USER_ID_KEY}'", e)
```
ivo-ritense: Not sure about this. If any implementation has some key of own authentication mechanism, like a server to server token, It will spam error in the log file.

Klaas-Ritense: I've changed it to a warning. Now, implementations that have their own authentication mechanism, can put the log level for this UserLoggingFilter to ERROR so that this line won't be logged

ivo-ritense (approved):

PR #1553, opened 2024-09-17
Bugfix: immutable camunda entities

theo-ritense:
### Describe the changes
Link to the related issue: TP 122345

### Breaking changes

### Documentation

> - 

- [ ] Yes
- [x] Not applicable

### Tests
- [ ] Yes
- [x] Not applicable

- [ ] Yes
- [x] Not applicable

### Security
- [ ] Yes
- [x] Not applicable

- [ ] Yes
- [x] Not applicable

- [ ] Yes
- [x] Not applicable

### Dependencies
- [ ] Yes
- [x] Not applicable

### DIFF
--- a/core/src/main/kotlin/com/ritense/valtimo/camunda/domain/CamundaExecution.kt
+++ b/core/src/main/kotlin/com/ritense/valtimo/camunda/domain/CamundaExecution.kt
@@ -27,78 +27,81 @@ import jakarta.persistence.ManyToOne
 import jakarta.persistence.OneToMany
 import jakarta.persistence.Table
 import jakarta.persistence.Transient
+import org.hibernate.annotations.Immutable
 
+@Immutable
 @Entity
 @Table(name = "ACT_RU_EXECUTION")
 @JsonIdentityInfo(generator = ObjectIdGenerators.PropertyGenerator::class, property = "id")
 class CamundaExecution(
 
     @Id
-    @Column(name = "ID_")
+    @Column(name = "ID_", insertable = false, updatable = false)
     val id: String,
 
-    @Column(name = "REV_")
+    @Column(name = "REV_", insertable = false, updatable = false)
     val revision: Int,
 
     @ManyToOne(fetch = FetchType.LAZY)
-    @JoinColumn(name = "ROOT_PROC_INST_ID_")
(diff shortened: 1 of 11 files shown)


Klaas-Ritense (approved):

core/src/main/kotlin/com/ritense/valtimo/camunda/domain/CamundaExecution.kt
```diff
@@ -27,78 +27,81 @@ import jakarta.persistence.ManyToOne
...
 
     @ManyToOne(fetch = FetchType.LAZY)
-    @JoinColumn(name = "ROOT_PROC_INST_ID_")
+    @JoinColumn(name = "ROOT_PROC_INST_ID_", insertable = false, updatable = false)
     val rootProcessInstance: CamundaExecution?,
 
     @ManyToOne(fetch = FetchType.LAZY)
-    @JoinColumn(name = "PROC_INST_ID_")
+    @JoinColumn(name = "PROC_INST_ID_", insertable = false, updatable = false)
     var processInstance: CamundaExecution?,
```
sdegroot, line 51: `var` implies that it is mutable? Is there a reason why this remains `var` ?

theo-ritense: @sdegroot I could find similar 2 usages of this `var`:
`execution.setProcessInstance(execution);`

I guess that could be changed to a custom getter returning `this` when the value is `null`, but that's out of scope for this PR. Remember: this is Camunda logic territory what we're trying to mimic.

sdegroot: I'd say that if you mark a class as `@Immutable`, it should be immutable :) 

This might cause problems later on. I can imagine hibernate is not happy if it notices that the property value is changed while it is set to `updatable = false` 

But I have not investigated the impact of having to change this. I'll leave it up to the team to decide

theo-ritense: The annotation name is a bit unfortunate:
https://docs.jboss.org/hibernate/orm/5.2/javadocs/org/hibernate/annotations/Immutable.html

`An immutable entity may not be updated by the application. Updates to an immutable entity will be ignored, but no exception is thrown.`

PR #1559, opened 2024-09-19
feature: logging

mbritense:
closes https://ritense.tpondemand.com/entity/118274-be-update-zgw-modules-to-use
https://ritense.tpondemand.com/entity/118272-be-implement-framework-for-logging-technical
https://ritense.tpondemand.com/entity/118275-be-storing-technical-logs https://ritense.tpondemand.com/entity/118278-be-periodically-cleaning-up-technical-logs https://ritense.tpondemand.com/entity/121634-be-retrieving-technical-logs-via-rest  https://ritense.tpondemand.com/entity/118277-be-searching-technical-logs


logging/src/main/kotlin/com/ritense/logging/service/LoggingEventDeletionService.kt
```diff
@@ -0,0 +1,53 @@
...
+@SkipComponentScan
+class LoggingEventDeletionService(
+    private val retentionInMinutes: Long,
+    private val loggingEventRepository: LoggingEventRepository,
+    private val loggingEventPropertyRepository: LoggingEventPropertyRepository,
+    private val loggingEventExceptionRepository: LoggingEventExceptionRepository,
+) {
+
+    @Scheduled(
+        fixedRateString = "\${valtimo.logging.retentionInMinutes:43200}",
```
marijnritense: Why do we only delete these once every 30 days? Shouldn't this job run more often than that?

logging/src/main/kotlin/com/ritense/logging/LoggingAutoConfiguration.kt
```diff
@@ -0,0 +1,114 @@
...
+    @Bean
+    @ConditionalOnMissingBean(LoggableResourceAspect::class)
+    fun loggableResourceAspect(): LoggableResourceAspect {
+        return LoggableResourceAspect()
+    }
+
+    @Bean
+    @ConditionalOnMissingBean(LoggingEventDeletionService::class)
+    fun loggingEventDeletionService(
+        @Value("\${valtimo.logging.retentionInMinutes:43200}") retentionInMinutes: Long,
```
marijnritense: Why is this set to 30 days instead of the 7 days that was mentioned in the story?

logging/src/main/kotlin/com/ritense/logging/service/LoggingEventDeletionService.kt
marijnritense: This should be configurable separately from the retention time.

marijnritense: Could be changed to support cron notation instead as this offers more flexibility

marijnritense: Run once a day

logging/src/main/kotlin/com/ritense/logging/LoggingAutoConfiguration.kt
marijnritense: Suggestion is to reduce this to 21 days

marijnritense (approved):

PR #1568, opened 2024-09-25
added logging

Tom-Ritense:


zgw/object-management/src/main/kotlin/com/ritense/objectmanagement/service/ObjectManagementFacade.kt
```diff
@@ -141,7 +146,7 @@ class ObjectManagementFacade(
         )
 
         try {
-            logger.trace { "Creating object $objectRequest" }
+            logger.info { "Creating object $objectRequest" }
```
ivo-ritense: I think we shouldn't want to log the whole `objectRequest` as info here because it might be large, complex and contain sensitive data.

zgw/object-management/src/main/kotlin/com/ritense/objectmanagement/service/ObjectManagementFacade.kt
```diff
@@ -167,7 +174,7 @@ class ObjectManagementFacade(
         )
 
         try {
-            logger.trace { "Updating object $objectRequest" }
+            logger.info { "Updating object $objectRequest" }
```
ivo-ritense: I think we shouldn't want to log the whole objectRequest as info here because it might be large, complex and contain sensitive data.

zgw/object-management/src/main/kotlin/com/ritense/objectmanagement/service/ObjectManagementFacade.kt
marijnritense: Is this a useful thing to log? I think this should be covered by the plugin action.

zgw/object-management/src/main/kotlin/com/ritense/objectmanagement/service/ObjectManagementFacade.kt
Tom-Ritense: I propose to remove the logger line here

zgw/object-management/src/main/kotlin/com/ritense/objectmanagement/service/ObjectManagementFacade.kt
Tom-Ritense: I propose to remove the logger line here

marijnritense (approved):

PR #1570, opened 2024-09-25
#123038 save mdc context for exception

Klaas-Ritense:


Tom-Ritense (commented):
Should we als cover the Cronjob and RabbitMQ event handler flows? So that each of those don't lose the MDC context in case of errors. @ivo-ritense @marijnritense

ivo-ritense (approved):

ivo-ritense:
> Should we als cover the Cronjob and RabbitMQ event handler flows? So that each of those don't lose the MDC context in case of errors. @ivo-ritense @marijnritense

I think we should

Klaas-Ritense:
@ivo-ritense @Tom-Ritense The errors in CronJobs now also have MDC information. This is done by the `loggingErrorWithContextThreadPoolTaskScheduler` bean.

We don't have a handler yet in Valtimo for RabbitMQ events

PR #1595, opened 2024-10-18
Make docker composeUpGzac part of bootRun

Klaas-Ritense:


app/gzac/build.gradle
```diff
@@ -29,6 +29,7 @@ dependencies {
 apply from: "gradle/publishing.gradle"
 
 bootRun {
+    dependsOn("composeUpGzac")
```
marijnritense: I'm wondering if this should not instead be a separate task, a bootRunWithCompose. That way those who prefer to use compose via the command line still have that as an option.

Additionally, from what I've gathered there can be issues with running composeUpGzac in general. This is an issue on MacOS if your docker binaries are not in the standard location due to java21 being unable to access the path variable. Also see the accompanying issues [here](https://github.com/avast/gradle-docker-compose-plugin/issues/435) and [here](https://youtrack.jetbrains.com/issue/IDEA-334183/Gradle-cannot-access-Environmental-Variables).

The workaround for this has already been added for NL Portal, [see here](https://github.com/nl-portal/nl-portal-backend-libraries/blob/20b21b6c19b0b30a47314fd99d12338d632df9ce/build.gradle.kts#L216). Maybe we could add this here as well part of this PR? Otherwise there will be people who are unable to contribute to Valtimo.

ivo-ritense (approved):

PR #1619, opened 2024-10-29
Link from dashboard

ivo-ritense:
Also introducing changelog checksum migration


changelog/src/main/kotlin/com/ritense/valtimo/changelog/service/ChangelogService.kt
```diff
@@ -52,19 +53,31 @@ class ChangelogService(
...
+                md5sum = md5sum,
+                checksumType = ChangesetCheckSumType.FILE_HASH
             )
         )
     }
 
     fun deleteChangesetsByKey(key: String?) = changesetRepository.deleteAllByKey(key)
 
-    fun isNewValidChangeset(changesetId: String, md5sum: String): Boolean {
+    fun isNewValidChangeset(changesetId: String, md5sum: String, legacyCheckSum: String): Boolean {
```
marijnritense, line 64: Technically a breaking change, do we want to worry about this or just accept this?

marijnritense (approved):

PR #1634, opened 2024-11-07
Story/provide documentid for starting processes for case

marijnritense:
Added document contexts to several places when the documentId is available; when retrieving process links, when completing a form associated, and when retrieving associated processes

### DIFF
--- a/authorization/src/main/kotlin/com/ritense/authorization/specification/AuthorizationSpecification.kt
+++ b/authorization/src/main/kotlin/com/ritense/authorization/specification/AuthorizationSpecification.kt
@@ -91,13 +91,12 @@ abstract class AuthorizationSpecification<T : Any>(
                 permission.appliesInContext(
                     relatedEntityAuthorizationRequest.context?.resourceType,
                     relatedEntityAuthorizationRequest.context?.entity
-                )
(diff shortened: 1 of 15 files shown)


authorization/src/main/kotlin/com/ritense/authorization/specification/AuthorizationSpecification.kt
```diff
@@ -86,6 +88,10 @@ abstract class AuthorizationSpecification<T : Any>(
                     && relatedEntityAuthorizationRequest.action == permission.action
             }
             .firstOrNull { permission ->
+                permission.appliesInContext(
```
ivo-ritense, line 91: Does this actually do anything? It doesn't seem to be returned or throw anything.

marijnritense: The `isAuthorizedForRelatedEntity` method returns true if a permission matches, and false if none matched.

ivo-ritense: But doesn't it do the match on the last statematch in the `firstOrNull` block?

ivo-ritense (approved):

PR #1637, opened 2024-11-08
Story/try catch on task list

floris-thijssen-ritense:
### Describe the changes
**Relevant comments:**
In Utrecht CommUtr, we have tasks that reference to users or documents that no longer exists.
This throws errors when opening the task screen, resulting in the entire task screen being unable to load.

This change doesn't throw the error all the way to the endpoints, but rather logs an error and skips when it fails to find a document/user, so that any tasks that can be built are still returned

### Breaking changes

- [X] The contribution only contains changes that are not breaking.

### Documentation

>

- [ ] Yes
- [ ] Not applicable

### Tests
- [ ] Yes
- [X] Not applicable

- [ ] Yes
- [X] Not applicable

### Security
- [ ] Yes
- [ ] Not applicable

- [ ] Yes
- [X] Not applicable

- [ ] Yes
- [X] Not applicable

### Dependencies
- [ ] Yes
- [X] Not applicable

### DIFF
--- a/core/src/main/java/com/ritense/valtimo/service/CamundaTaskService.java
+++ b/core/src/main/java/com/ritense/valtimo/service/CamundaTaskService.java
@@ -86,6 +86,7 @@ import java.time.LocalDateTime;
 import java.util.Collection;
 import java.util.List;
 import java.util.Map;
+import java.util.Objects;
 import java.util.Optional;
 import java.util.Set;
 import java.util.UUID;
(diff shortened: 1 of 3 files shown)


core/src/main/java/com/ritense/valtimo/service/CamundaTaskService.java
```diff
@@ -366,34 +371,40 @@ public Page<TaskExtended> findTasksFiltered(
...
+                        executionId,
+                        businessKey,
+                        processDefinitionId,
+                        processDefinitionKey,
+                        valtimoUser,
+                        context
+                    );
+                } catch(Exception e) {
+                    logger.error("Failed to build task "+tuple, e);
+                    return null;
```
marijnritense: Rather than filtering out the results, I think it makes more sense to update those records so the users are asynchronously unassigned from the relevant tasks. Additionally, the user should just be excluded from the results, but the task itself should still be returned.

floris-thijssen-ritense: It does not seem like a good idea to us to remove the users from the tasks, simply because you tried to retrieve a list of tasks. This would mean throwing away data that could not be retrieved anymore, and should be done with some consideration, not as a side effect.

For instance, a bug where the migration from userEmail to userId fails, would end up in wiping all the users from all the tasks.

ivo-ritense (approved):

PR #1638, opened 2024-11-08
added param to interface so that it can accept the current page of the formio wizard

laurens-ritense:
…e formio wizard

### Describe the changes
Link to the related Github issue: 

**Relevant comments:**
> 

### Breaking changes

### Documentation

- [x] Release notes have been written for these changes. 

Link to the pull request in the [Valtimo documentation repository](https://github.com/valtimo-platform/valtimo-documentation): https://github.com/valtimo-platform/valtimo-documentation/pull/754
>

- [x] Yes
- [ ] Not applicable

### Tests
- [ ] Yes
- [x] Not applicable

- [ ] Yes
- [x] Not applicable

### Security
- [ ] Yes
- [x] Not applicable

- [ ] Yes
- [x] Not applicable

- [ ] Yes
- [x] Not applicable

### Dependencies
- [ ] Yes
- [x] Not applicable

### DIFF
--- a/form-view-model/src/main/kotlin/com/ritense/formviewmodel/service/FormViewModelService.kt
+++ b/form-view-model/src/main/kotlin/com/ritense/formviewmodel/service/FormViewModelService.kt
@@ -88,7 +88,7 @@ class FormViewModelService(
         val viewModelLoader = viewModelLoaderFactory.getViewModelLoader(formName) ?: return null
         val viewModelType = viewModelLoader.getViewModelType()
         if (isWizard == true) {
-            return parseViewModel(submission, viewModelType).update(page = page)
+            return parseViewModel(submission, viewModelType).update(task = task, page = page)
         }
         return parseViewModel(submission, viewModelType).update(task)
     }
--- a/form-view-model/src/main/kotlin/com/ritense/formviewmodel/service/FormViewModelService.kt
+++ b/form-view-model/src/main/kotlin/com/ritense/formviewmodel/service/FormViewModelService.kt
@@ -61,7 +61,7 @@ class FormViewModelService(
     fun updateStartFormViewModel(
         formName: String,
         processDefinitionKey: String,
-        page: Int?,
+        page: Int? = null,
         submission: ObjectNode
     ): ViewModel? {
         processAuthorizationService.checkAuthorization(processDefinitionKey)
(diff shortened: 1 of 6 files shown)


form-view-model/src/main/kotlin/com/ritense/formviewmodel/service/FormViewModelService.kt
```diff
@@ -60,18 +60,20 @@ class FormViewModelService(
 
     fun updateStartFormViewModel(
         formName: String,
-        submission: ObjectNode,
-        processDefinitionKey: String
+        processDefinitionKey: String,
+        page: Int?,
```
marijnritense: All additions of the page argument (apart from the REST API one) are breaking for java, as all arguments have to be provided for java. Can you instead introduce a second variant of these that work the same way the used to, and mark those as deprecated?

laurens-ritense: I tried something else that should make it non breaking. Have a look and tell me what you think.

marijnritense: Still a breaking change, you're adding in multiple arguments that weren't mandatory before. On top of that, you've changed the ordering around so unless people are using named arguments.

laurens-ritense: This method is not used directly by implementations and is only triggerd automatically by the endpoint the only change for someone using FVM is in the ViewModel interface which doesn't have a breaking change.

laurens-ritense: If an implementation developer triggers this method manually they are doing something very wrong.

form-view-model/src/main/kotlin/com/ritense/formviewmodel/service/FormViewModelService.kt
```diff
@@ -58,20 +58,54 @@ class FormViewModelService(
         return viewModelLoaderFactory.getViewModelLoader(formName)?.load(task)
     }
 
+    @Deprecated("Use method with page support instead")
     fun updateStartFormViewModel(
         formName: String,
-        submission: ObjectNode,
-        processDefinitionKey: String
+        processDefinitionKey: String,
```
marijnritense, line 64: Can you fix this so the order of arguments is still the same?

marijnritense (approved):

PR #1642, opened 2024-11-15
story: additional info tasks

mbritense:
closes https://ritense.tpondemand.com/entity/127954-febe-additional-info-tasks

### DIFF
--- a/app/gzac/src/main/resources/config/processlink/bezwaar.processlink.json
+++ b/app/gzac/src/main/resources/config/processlink/bezwaar.processlink.json
@@ -11,7 +11,11 @@
         "formDefinitionName": "user-task-validate-request",
         "processLinkType": "form",
         "formDisplayType": "panel",
-        "formSize": "medium"
+        "formSize": "medium",
+        "subtitles": [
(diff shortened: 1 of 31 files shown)


app/gzac/src/main/resources/config/processlink/bezwaar.processlink.json
```diff
@@ -11,7 +11,11 @@
         "formDefinitionName": "user-task-validate-request",
         "processLinkType": "form",
         "formDisplayType": "panel",
-        "formSize": "medium"
+        "formSize": "medium",
+        "subtitles": [
```
Klaas-Ritense, line 15: I wonder if this breaks the hash. But I think it was fixed by Ivo. Can you check with him?

ivo-ritense: Normally the change to this specific file would break the hash, but with my change the DTO being changed shouldn't. But this is not a changelog import file so it doesn't matter either way.

Klaas-Ritense (approved):

PR #1673, opened 2024-12-10
#57: Configurable Documenten API Upload fields

### LINKED ISSUE #57: Prefill and hide document metadata fields
## Why
*Why do we need this feature?*

- **Problem Statement**: What problem does this feature solve? What are the underlying pain points that make this feature necessary? 
This feature improves the user experience when uploading a document and filling in the required metadata fields. When a case handle has to upload a lot of files, filling in the metadata for each individual file can take a lot of time. The time needed can be reduced by configuring which fields should be filled in by the case handler, and which fields have a default or fixed value.
- **Value Proposition**: How will this feature add value to the users or the product? Why is it important to solve this problem?
This feature adds value because it reduces the time needed by the case handler to upload documents.

## How
*How do we envision this feature working?*

- **Guiding Principles**: What are the guiding principles or criteria for this feature? Consider user experience, design consistency, and alignment with the project's goals.
Guiding principle: don't ask the user for input while they shouldn't have a choice. 
- **Key Use Cases**: What are the primary scenarios in which this feature will be used? Describe the user journey and how this feature fits into it.
See above.

## What
*What needs to be done to implement this feature?*

- **Feature Description**: Provide a detailed description of the feature. What are the main components or elements?
The system administrator is able to configure default values for each of the document metadata fields. Per field, the system admin can configure if the field should be open to fill in, should have a default value, should be editable and should be visible.
- **Requirements**: List any specific requirements or constraints. Consider technical, design, or performance requirements.
The metadata field configuration should be configurable per case-type. The configuration should be auto-deployable and should be part of the case definition import and export.
- **Acceptance Criteria**: Define the acceptance criteria for this feature. What conditions must be met for this feature to be considered complete?
See above.

## Additional Context
- **Stakeholders**: Who are the key stakeholders or contributors for this feature?
- **References**: Include any relevant links, documents, or resources that provide additional context or support for this feature.
- **Potential Challenges**: Identify any potential challenges or obstacles that might arise during the implementation of this feature.

---

Klaas-Ritense:


zgw/documenten-api/src/main/kotlin/com/ritense/documentenapi/web/rest/DocumentenApiManagementResource.kt
```diff
@@ -117,4 +121,23 @@ class DocumentenApiManagementResource(
...
+    @GetMapping("/v1/case-definition/{caseDefinitionName}/zgw-document/upload-field")
+    fun getUploadFields(
+        @LoggableResource("documentDefinitionName") @PathVariable(name = "caseDefinitionName") caseDefinitionName: String,
+    ): ResponseEntity<List<DocumentenApiUploadFieldDto>> {
+        val list = documentenApiService.getUploadFields(caseDefinitionName)
+        return ResponseEntity.ok(list.map { DocumentenApiUploadFieldDto.of(it) })
+    }
+
+    @RunWithoutAuthorization
+    @PutMapping("/v1/case-definition/{caseDefinitionName}/zgw-document/upload-field")
```
ivo-ritense, line 135: should this not be a mangement url?

app/gzac/src/main/resources/config/zgw/bezwaar.zgw-document-upload-fields.json
```diff
@@ -0,0 +1,49 @@
+{
```
ivo-ritense, line 1: Is this the format of the file you envision? i would think that the fields should be in the same array since they are for the same case definition.

ivo-ritense (approved):

app/gzac/src/main/resources/config/zgw/bezwaar.zgw-document-upload-fields.json
Klaas-Ritense: This is the format described in the solution design

PR #1679, opened 2024-12-17
story: task data source dashboard

mbritense:
closes https://ritense.tpondemand.com/entity/130704-befe-task-data-on-dashboards


contract/src/main/kotlin/com/ritense/valtimo/contract/authorization/UserManagementServiceHolder.kt
```diff
@@ -14,7 +14,7 @@
  * limitations under the License.
  */
 
-package com.ritense.authorization
+package com.ritense.valtimo.contract.authorization
```
Klaas-Ritense, line 17: This class and the other classes don't belong in the contract module

mbritense: I discussed this with @ivo-ritense, there was no other reasonable option it seems

contract/src/main/kotlin/com/ritense/valtimo/contract/dashboard/QueryCondition.kt
```diff
@@ -0,0 +1,147 @@
...
+import org.springframework.expression.spel.support.StandardEvaluationContext
+import java.time.LocalDateTime
+
+data class QueryCondition<T : Comparable<T>>(
+    val queryPath: String,
+    val queryOperator: ExpressionOperator,
+    @JsonDeserialize(using = ComparableDeserializer::class)
+    val queryValue: T
+) {
+    private fun <T> queryValueIsNull(target: T): Boolean {
```
ivo-ritense, line 78: Generally you would put the private methods at the bottom of the class.

contract/src/main/kotlin/com/ritense/valtimo/contract/dashboard/QueryCondition.kt
```diff
@@ -0,0 +1,147 @@
...
+        root: Root<*>,
+        criteriaBuilder: CriteriaBuilder,
+        pathExpressionFunction: (Class<Any>, String, Root<*>, CriteriaBuilder) -> Expression<Any>
+    ): Predicate {
+        pathExpressionFunction as (Class<T>, String, Root<*>, CriteriaBuilder) -> Expression<T>
+
+        if (queryValueIsDateTimeSpelExpression(queryValue)) {
+            return getPredicateFromDateTimeSpelExpression(
+                root,
+                this as QueryCondition<String>,
```
ivo-ritense: Why do we need to pass this in if it is always already the current object and the method is in the object?

core/src/test/kotlin/com/ritense/valtimo/dashboard/TaskWidgetDataSourceIntTest.kt
```diff
@@ -0,0 +1,204 @@
...
+        )
+
+        val result = taskWidgetDataSource.getTaskCount(dataSourceProperties)
+
+        assertThat(result.value).isEqualTo(defaultTasks.toLong())
+        assertThat(result.total).isEqualTo(3.toLong())
+    }
+
+    @Test
+    fun `should filter tasks created within the last minute`() {
```
ivo-ritense: I think when testing a filter you should not only have an entry that maatches the filter, but also one that doesn't.

contract/src/main/kotlin/com/ritense/valtimo/contract/authorization/UserManagementServiceHolder.kt
ivo-ritense: I think it's hard to get around some of this. it's a choice between having classes like this in contract, or having core be dependent on dashboard, which also isn't great. 

The classes that are in contract now are either general configuration classes, or classes that are meant to be re-usable between data sources.

core/src/test/kotlin/com/ritense/valtimo/dashboard/TaskWidgetDataSourceIntTest.kt
ivo-ritense: We talked about this and it's hard to mock camunda data so we will skip it in this case.

ivo-ritense (approved):

PR #1680, opened 2024-12-17
Verzoek mapping now accepts prefix 'object:'

### LINKED ISSUE #99: Allow acces to the meta data in the verzoek plugin
## Why
*Why do we need this feature?*
I want to be able to acces the meta data form a verzoek so i can use that data in my process

- **Problem Statement**: I want to have access to the meta data of a verzoek so i can use that data in my process/case

## How
*How do we envision this feature working?* 
I would suggest to use a verzoek: and there you can acces it in the mapping of u

- **Key Use Cases**: 
- A bsn does not have to be duplicated in the data within a verzoek
- You can access the reference of open forms of there submission

## What
*What needs to be done to implement this feature?*

- **Feature Description**: The feature will allow users to access metadata from a verzoek in a structured format. The metadata will be integrated into the existing mapping functionality.
- **Requirements**: You can acces the meta data in the mapping part of the verzoek plugin
- **Acceptance Criteria**: Users can view and select metadata from a verzoek in the mapping interface.
Metadata can be successfully integrated into workflows and cases.

## Additional Context
- **Stakeholders**: Ritense/den Haag
- **Potential Challenges**: -

---

Klaas-Ritense:


zgw/verzoek/src/main/kotlin/com/ritense/verzoek/VerzoekPluginEventListener.kt
```diff
@@ -33,24 +33,29 @@ import com.ritense.logging.withLoggingContext
...
 import mu.KotlinLogging
 import org.springframework.context.event.EventListener
+import org.springframework.stereotype.Component
 import org.springframework.transaction.annotation.Transactional
 import java.net.URI
 import com.ritense.processdocument.resolver.DocumentJsonValueResolverFactory.Companion.PREFIX as DOC_PREFIX
 import com.ritense.valueresolver.ProcessVariableValueResolverFactory.Companion.PREFIX as PV_PREFIX
 
+@SkipComponentScan
+@Component
```
ivo-ritense, line 56: You can also use `@AllOpen` i think

zgw/verzoek/src/main/kotlin/com/ritense/verzoek/VerzoekPluginEventListener.kt
```diff
@@ -124,6 +144,17 @@ open class VerzoekPluginEventListener(
         }
     }
 
+    protected fun getVerzoekDataFromPath(verzoekObject: JsonNode, path: String): JsonNode {
+        return if (path.startsWith("object:")) {
+            verzoekObject.at(path.substringAfterLast("object:"))
+        } else {
+            val verzoekDataData = verzoekObject["record"]["data"]["data"] ?: throw NotificatiesNotificationEventException(
+                "VerzoekObject /record/data/data is empty! For verzoek with type '${verzoekObject["type"]}'"
```
ivo-ritense:
```suggestion
                "VerzoekObject /record/data/data is cannot be found! For verzoek with type '${verzoekObject["type"]}'"
```

ivo-ritense: technically if it exists and is empty i don't think this found return null

ivo-ritense (dismissed):

marijnritense (approved):

PR #1688, opened 2025-01-09
Fix process-link type update for autodeployment

theo-ritense:
Updating a process-link to a different type via autodeployment was not possible because the mapper was selected by the current `processLinkType`.


marijnritense (commented):
Can you add a test to make sure this works as expected?

ivo-ritense (approved):

PR #1698, opened 2025-01-16
open tasks fails if a single permission fails

marijnritense:


process-link/src/main/kotlin/com/ritense/processlink/service/ProcessLinkActivityService.kt
```diff
@@ -61,10 +61,14 @@ class ProcessLinkActivityService(
...
-        val task = taskService.findTaskOrThrow(
-            byId(taskId.toString())
-                .and(byActive())
-        )
+        val task = try {
+            taskService.findTaskOrThrow(
+                byId(taskId.toString())
+                    .and(byActive())
+            )
+        } catch (exception: Exception) {
```
ivo-ritense, line 69: Is this type correct? This would mean that any possible error (which admittedly i can't think of many) would be translated to process link not found.

marijnritense: Fixed it based on our discussion

ivo-ritense (approved):

PR #1706, opened 2025-01-24
Replaced RequestScopeUserCache with an eviction policy based UserCache

maarten-ritense:
The RequestScopeUserCache introduces issues incase there is not Request scope available. So replacing the request scope user cache with a user cache which makes use of an eviction policy and is independent from the request scope.


Klaas-Ritense:
@maarten-ritense It seems we both made a fix for the same issue: https://github.com/valtimo-platform/valtimo-backend-libraries/pull/1704

PR #1714, opened 2025-01-29
Option to configure an external start case form per case

maarten-ritense:
Extended the case definition configuration so it offers the ability to configure if the case uses an external form for starting a case.


case/src/main/kotlin/com/ritense/case/domain/CaseDefinitionSettings.kt
```diff
@@ -20,17 +20,22 @@ import jakarta.persistence.Column
...
+    @Column(name = "case_definition_name", nullable = false, length = 256)
     val name: String,
-    @Column(name = "can_have_assignee")
+    @Column(name = "can_have_assignee", nullable = false)
     val canHaveAssignee: Boolean = false,
-    @Column(name = "auto_assign_tasks")
+    @Column(name = "auto_assign_tasks", nullable = false)
     val autoAssignTasks: Boolean = false,
+    @Column(name = "has_external_create_case_form", nullable = false)
+    val hasExternalCreateCaseForm: Boolean = false,
```
ThomasMinkeRitense: @maarten-ritense any reason why it's not called 'hasExternalStartForm'?

ivo-ritense (approved):

David-Kooi-Ritense (approved):

PR #1719, opened 2025-02-03
Bring back UserCache 

Klaas-Ritense:


keycloak-iam/src/main/java/com/valtimo/keycloak/service/CacheManagerUserCache.kt
```diff
@@ -0,0 +1,60 @@
...
+        val newValue = requestFunction(key)
+        if (newValue != null && !cacheType.cachedClass.java.isAssignableFrom(newValue!!::class.java)) {
+            throw IllegalArgumentException("The type of the value returned by the request function (${newValue!!::class.java.name}) does not match the cache type (${cacheType.cachedClass.java.name})")
+        }
+        cache.put(key, newValue)
+        return newValue
+    }
+
+    @CacheEvict(allEntries = true, value = ["EMAIL_ManageableUser", "USER_IDENTIFIER_ValtimoUser"])
+    @Scheduled(fixedRateString = "\${valtimo.keycloak.cache.ttl:PT1H}")
```
theo-ritense: This will result in irregular cleanups if the cachemanager is backed by a redis or other service with multiple instances. Effectively, this is more a `maxTtl` than a `ttl`, since the cache could be cleared after a 1 minute if the other instance started 1 minute later.

On the other hand, if you added Shedlock when the cachemanager is NOT backed by redis, it could result in the Cachemanager to be cleared on just one instance, so that is not a solution.

I suggest renaming the setting to `maxTtl`.

Klaas-Ritense: Renamed

keycloak-iam/src/main/java/com/valtimo/keycloak/service/KeycloakUserManagementService.java
```diff
@@ -61,16 +58,16 @@ public class KeycloakUserManagementService implements UserManagementService {
...
     public KeycloakUserManagementService(
         KeycloakService keycloakService,
         String keycloakClientName,
-        CacheManager cacheManager
+        UserCache userCache
     ) {
         this.keycloakService = keycloakService;
         this.clientName = keycloakClientName;
-        this.cacheManager = cacheManager;
+        this.userCache = userCache;
```
marijnritense, line 70: Caught between a rock and a hard place here I think, but now this is breaking too.

marijnritense (approved):

PR #1723, opened 2025-02-13
doc: Add PBAC permissions to documentenAPI documents

Klaas-Ritense:


app/gzac/src/main/resources/config/application.yml
```diff
@@ -250,6 +250,8 @@ valtimo:
             enabled: true
     logging:
         retentionInMinutes: 1440
+    featureToggles:
```
ivo-ritense: It makes sense i think to put this together with the other ac feature toggle:

```
valtimo:
    authorization:
        dashboard:
            enabled: true
```

ivo-ritense (dismissed):

app/gzac/src/main/resources/config/application.yml
Klaas-Ritense: renamed

ivo-ritense (approved):

PR #1740, opened 2025-02-26
Bugfix: Will no longer upload a document that has already been uploaded

Klaas-Ritense:
Error scenario:
1. Open a task with an upload form
2. Upload a document
3. The resulting tempId is saved in the document-json-schema. This is default behavior
4. Open a task with an upload field with the same key. This upload form is now prefilled with the tempId
5. Complete task without changing anything
6. ERROR: Temp file no longer exists

The fix:
We have permanent metadata for a tempFileId. Whit this metadata we track if the document was already uploaded and linked.


ivo-ritense (approved):

David-Kooi-Ritense (approved):
approving based on test results. 
2nd unintentional upload seems to be resolved.

PR #1749, opened 2025-03-07
Set BPMN service task expression to ${null}

Klaas-Ritense:


core/src/main/java/com/ritense/valtimo/service/CamundaProcessService.java
```diff
@@ -323,6 +337,70 @@ private void setProcessesExecutable(BpmnModelInstance bpmnModel) {
...
+            .filter(sendEvent -> sendEvent.getId().matches("Event_[a-z0-9]{7}"))
+            .flatMap(sendEvent -> sendEvent.getChildElementsByType(MessageEventDefinition.class).stream())
+            .forEach(event -> {
+                if (event.getCamundaType() == null
+                    && event.getCamundaClass() == null
+                    && event.getCamundaExpression() == null
+                    && event.getCamundaDelegateExpression() == null) {
+                    String messageName = event.getMessage() == null ? "MY_MESSAGE" : event.getMessage().getName();
+                    event.setCamundaExpression(
+                        "${correlationService.sendMessageToAll(\"" + messageName + "\",execution)}"
```
theo-ritense:
```suggestion
                        "${correlationService.sendMessageToAll(\"" + messageName + "\", execution)}"
```

core/src/main/java/com/ritense/valtimo/service/CamundaProcessService.java
```diff
@@ -323,6 +337,70 @@ private void setProcessesExecutable(BpmnModelInstance bpmnModel) {
...
+                    );
+                }
+            });
+    }
+
+    private void setToRoleUserWhenUserTaskCandidateGroupsIsEmpty(BpmnModelInstance bpmnModel) {
```
ThomasMinkeRitense: @Klaas-Ritense why would we do this?

core/src/main/java/com/ritense/valtimo/service/CamundaProcessService.java
```diff
@@ -323,6 +337,70 @@ private void setProcessesExecutable(BpmnModelInstance bpmnModel) {
...
+        bpmnModel.getModelElementsByType(UserTask.class).forEach(userTask -> {
+            if (userTask.getId().matches("Activity_[a-z0-9]{7}")
+                && userTask.getCamundaCandidateUsers() == null
+                && userTask.getCamundaCandidateGroups() == null) {
+                userTask.setCamundaCandidateGroups("ROLE_USER");
+            }
+        });
+    }
+
+    private void setTo60SecondsWhenTimerIsEmpty(BpmnModelInstance bpmnModel) {
```
ThomasMinkeRitense, line 410: I think this goes too far, a timer event really should be configured explicitly by the user. Forgetting to put in a 4 week duration can't mean you get one minute by default.

ThomasMinkeRitense (dismissed):

core/src/main/java/com/ritense/valtimo/service/CamundaProcessService.java
Klaas-Ritense: @ThomasMinkeRitense 

The goal of this PR is to streamline and speed up BPMN modeling, especially for new users. Setting this property by default addresses several common pain points:

- **Prevents users from forgetting to set the property**: This is a field that’s often overlooked, leading to incomplete configurations.
- **Simplifies the setup**: By setting this default, it reduces the number of properties users need to manually configure before the BPMN works properly.
- **Improves the user experience**: Drag-and-drop functionality for User-tasks will work out of the box without additional configuration, making the tool more intuitive for beginners.
- **Prevents common mistakes**: Some users mistakenly enter roles in the `CandidateUsers` field instead of `CandidateGroups`. This default helps avoid that issue.

Overall, this change aligns with the objective of improving usability and efficiency, particularly for those less familiar with BPMN modeling.

core/src/main/java/com/ritense/valtimo/service/CamundaProcessService.java
Klaas-Ritense: @ThomasMinkeRitense 

This property is mandatory, and if it's not entered, it will trigger an error. By assigning a default value to this field, we can significantly enhance the user experience by preventing errors and making the process smoother for users.

core/src/main/java/com/ritense/valtimo/service/CamundaProcessService.java
ThomasMinkeRitense: @Klaas-Ritense I totally understand that, and I think that it's a great improvement for the service task implementation + expression. But that's only because we're pretty confident that this is the right default value. For timers, there is no logical default value and we shouldn't make it easier to make configuration mistakes. 

As for the candidate users: it's not a required field, why would we even set it to ROLE_USER?

Klaas-Ritense: @ThomasMinkeRitense 

Preventing configuration mistakes should be addressed through a thorough BPMN review process, rather than relying on a single person to get the configuration correct. 

The default values are intended to help new users get started quickly and avoid common pitfalls.

These are two distinct topics with different solutions

ivo-ritense (approved):

ivo-ritense (approved):

ThomasMinkeRitense (approved):

PR #1750, opened 2025-03-10
Story/case tags 132892

erik-ritense:
### Describe the changes
Link to the related Github issue: https://github.com/generiekzaakafhandelcomponent/gzac-issues/issues/93

**Relevant comments:**
> First many-to-many relation for json_schema_document

### Breaking changes

### Documentation

>

- [x] Yes
- [ ] Not applicable

### Tests
- [x] Yes
- [ ] Not applicable

- [x] Yes
- [ ] Not applicable

### Security
- [x] Yes
- [ ] Not applicable

- [x] Yes
- [ ] Not applicable

- [x] Yes
- [ ] Not applicable

### Dependencies
- [ ] Yes
- [x] Not applicable


app/gzac/src/main/resources/bpmn/bezwaar.bpmn
```diff
@@ -1,80 +1,107 @@
...
-    <bpmn:sequenceFlow id="Flow_162004p" sourceRef="start-event-bezwaar-proces-gestart" targetRef="validate-request" />
     <bpmn:startEvent id="start-event-bezwaar-proces-gestart" name="Bezwaar proces gestart">
       <bpmn:outgoing>Flow_162004p</bpmn:outgoing>
     </bpmn:startEvent>
-    <bpmn:sequenceFlow id="Flow_0ytx5d7" sourceRef="validate-request" targetRef="in-behandeling" />
+    <bpmn:serviceTask id="Activity_0r8lucn" name="add tag 1" camunda:expression="${documentDelegateService.addCaseTag(execution,&#34;test1&#34;)}">
+      <bpmn:incoming>Flow_162004p</bpmn:incoming>
+      <bpmn:outgoing>Flow_1e5lhmc</bpmn:outgoing>
+    </bpmn:serviceTask>
+    <bpmn:serviceTask id="Activity_1eiexpp" name="add tag 2" camunda:expression="${documentDelegateService.addCaseTag(execution,&#34;test2&#34;)}">
```
ivo-ritense, line 32: This would be nice to do through a process link. Maybe an improvement for later?

document/src/main/java/com/ritense/document/domain/impl/JsonSchemaDocument.java
```diff
@@ -363,6 +389,15 @@ public String internalStatus() {
         }
     }
 
+    @Override
+    public Set<CaseTagResponseDto> caseTags() {
+        if (caseTags == null) {
+            return null;
```
ivo-ritense: I think returning an empty set might make more sense.

document/src/main/java/com/ritense/document/domain/Document.java
```diff
@@ -45,6 +46,9 @@ public interface Document {
     @JsonProperty
     String internalStatus();
 
+    @JsonProperty
+    Set<CaseTagResponseDto> caseTags();
```
ivo-ritense: Not sure about the name of the method. This might imply to me that you get the domain objects instead of the DTOs. You can tell from the signature however. What do you think?

document/src/main/java/com/ritense/document/domain/search/AdvancedSearchRequest.java
```diff
@@ -86,6 +86,19 @@ public AdvancedSearchRequest statusFilter(String status) {
...
 
+    public Set<String> getCaseTagsFilter() {
+        return caseTagsFilter;
+    }
+
+    public void setCaseTagsFilter(Set<String> caseTagsFilter) {
+        this.caseTagsFilter = caseTagsFilter != null ? caseTagsFilter : new HashSet<>();
+    }
+
+    public AdvancedSearchRequest caseTagsFilters(String tag) {
```
ivo-ritense: You add 1 tag to the filter. Maybe the method name should reflect that. Something like `caseTagsFilter` or `addCaseTagsFilter`

document/src/main/java/com/ritense/document/domain/search/AdvancedSearchRequest.java
```diff
@@ -26,8 +26,8 @@ public class AdvancedSearchRequest {
     private SearchOperator searchOperator = SearchOperator.AND;
     private AssigneeFilter assigneeFilter = AssigneeFilter.ALL;
     private List<OtherFilter> otherFilters = List.of();
-
     private Set<String> statusFilter = new HashSet<>();
+    private Set<String> caseTagsFilter = new HashSet<>();
```
ivo-ritense, line 30: It's probably not worth it right now, but we could consider wrapping this in an object. If in the future we want to make it possible to search case tags based on AND and OR we could add that more easily. Could also be done by having another argument on root level however.

document/src/main/java/com/ritense/document/service/DocumentService.java
```diff
@@ -73,6 +73,8 @@ public interface DocumentService {
 
     void setInternalStatus(Document.Id documentId, @Nullable String internalStatusKey);
 
+    void addCaseTag(Document.Id documentId, String caseTagKey);
```
ivo-ritense, line 76: Can we only add them, and not remove?

document/src/main/java/com/ritense/document/service/impl/JsonSchemaDocumentService.java
```diff
@@ -734,6 +739,40 @@ public void setInternalStatus(
...
+        var caseTag = caseTagKey != null ? caseTagService.get(
+            document.definitionId().name(),
+            caseTagKey
+        ) : null;
+        document.addCaseTag(caseTag);
+
+        documentRepository.save(document);
+
+        outboxService.send(() ->
+            new DocumentStatusChanged(
```
ivo-ritense: Is this the correct event for this method?

document/src/main/kotlin/com/ritense/document/domain/CaseTagId.kt
```diff
@@ -0,0 +1,46 @@
...
+) : AbstractId<CaseTagId>() {
+    init {
+        require(caseDefinitionName.isNotBlank()) { "caseDefinitionName was blank!" }
+        require(key.isNotBlank()) { "key was blank!" }
+    }
+
+    companion object {
+        @JvmStatic
+        fun of(caseDefinitionName: String, key: String?): CaseTagId? {
+            return if (key == null) {
```
ivo-ritense: Not sure why the key should be nullable?

document/src/main/kotlin/com/ritense/document/exception/CaseTagAlreadyExistsException.kt
```diff
@@ -0,0 +1,24 @@
...
+ * limitations under the License.
+ */
+
+package com.ritense.document.exception
+
+class CaseTagAlreadyExistsException(
+    caseTagKey: String,
+    caseDefinitionName: String
+) : RuntimeException(
+    "A case tag with key $caseTagKey already exists for case with key $caseDefinitionName"
```
ivo-ritense: maybe to be complete and unambiguous:

```suggestion
    "A case tag with key $caseTagKey already exists for case definition with key $caseDefinitionName"
```

document/src/main/kotlin/com/ritense/document/exception/CaseTagNotFoundException.kt
```diff
@@ -0,0 +1,24 @@
...
+ * limitations under the License.
+ */
+
+package com.ritense.document.exception
+
+class CaseTagNotFoundException(
+    caseTagKey: String,
+    caseDefinitionName: String
+) : RuntimeException(
+    "A case tag with key $caseTagKey could not be found for case with key $caseDefinitionName"
```
ivo-ritense: So above
```suggestion
    "A case tag with key $caseTagKey could not be found for case definition with key $caseDefinitionName"
```

document/src/main/kotlin/com/ritense/document/exporter/CaseTagExporter.kt
```diff
@@ -0,0 +1,61 @@
...
+    override fun supports() = DocumentDefinitionExportRequest::class.java
+
+    override fun export(request: DocumentDefinitionExportRequest): ExportResult {
+        val caseTags = caseStatusService.getCaseTags(request.name)
+
+        if (caseTags.isEmpty()) {
+            return ExportResult()
+        }
+
+        val caseTabChangeset = CaseTagChangeset(
```
ivo-ritense:
```suggestion
        val caseTagChangeset = CaseTagChangeset(
```

document/src/main/kotlin/com/ritense/document/service/CaseTagService.kt
```diff
@@ -0,0 +1,150 @@
...
+        return caseTagRepository.getReferenceById(CaseTagId(caseDefinitionName, caseTagKey))
+    }
+
+    fun exists(caseDefinitionName: String, caseTagKey: String): Boolean {
+        return caseTagRepository.existsByIdCaseDefinitionNameAndIdKey(caseDefinitionName, caseTagKey)
+    }
+
+    fun create(
+        caseDefinitionName: String,
+        @Valid request: CaseTagCreateRequestDto) : CaseTag{
```
ivo-ritense: When the arguments span multiple lines, we usually put the `) : CaseTag {` on a new line. Also, please use a space between the return type and brace to open the method.

document/src/main/resources/config/liquibase/changelog/20250210-add-case-tag-link-table.xml
```diff
@@ -0,0 +1,63 @@
...
+        <createIndex indexName="idx_case_tag_link_json_schema_document_id" tableName="case_tag_link">
+            <column name="json_schema_document_id"/>
+        </createIndex>
+
+        <createIndex indexName="idx_case_tag_link_case_tag_id" tableName="case_tag_link">
+            <column name="case_tag_key"/>
+            <column name="case_definition_name"/>
+        </createIndex>
+
+        <addUniqueConstraint tableName="case_tag_link"
```
ivo-ritense: Is this needed if the primary key uses the same columns?

document/src/main/java/com/ritense/document/domain/Document.java
erik-ritense: For the method name, the same approach as for internalStatus() is used, in order to be consistent. Let's discuss..

document/src/main/java/com/ritense/document/domain/impl/JsonSchemaDocument.java
erik-ritense: Agreed and changed

document/src/main/java/com/ritense/document/domain/search/AdvancedSearchRequest.java
erik-ritense: I have thought about searching for AND as well. This can be a next step.

document/src/main/java/com/ritense/document/domain/search/AdvancedSearchRequest.java
erik-ritense: I will rename it to addCaseTagToFilter . Let's discuss if this method is even needed. It's not being used. It is only there to be consistent with internal case status (see line 84)

document/src/main/java/com/ritense/document/service/DocumentService.java
erik-ritense: Indeed, I think we should add a method for removal.

document/src/main/java/com/ritense/document/service/impl/JsonSchemaDocumentService.java
erik-ritense: No it is not correct. Should something like a DocumentCaseTagAdded event be introduced? Do we always need to send an event?

document/src/main/kotlin/com/ritense/document/domain/CaseTagId.kt
erik-ritense: Not sure as well, this class is based on InternalCaseStatusId which uses the same approach

document/src/main/kotlin/com/ritense/document/exporter/CaseTagExporter.kt
erik-ritense: fixed

document/src/main/kotlin/com/ritense/document/service/CaseTagService.kt
erik-ritense: fixed

document/src/main/resources/config/liquibase/changelog/20250210-add-case-tag-link-table.xml
erik-ritense: yes, indeed. I am removing the addUniqueConstraint.

document/src/main/java/com/ritense/document/service/DocumentService.java
erik-ritense: The removeCaseTag method has been added. I will add tests as well

document/src/main/java/com/ritense/document/service/impl/JsonSchemaDocumentService.java
erik-ritense: As for the DocumentStatusChanged as it is being used for internal case status, it looks to me that only the json content is being send. Let's discuss how much sense this makes, while the change of internal case status (and for this PR the case tag) is outside of the json content.

erik-ritense: Nevermind, I now see that DocumentStatusChanged does contain the meta data of the document.

erik-ritense: This has been fixed by introducing a DocumentTagsChanged event

David-Kooi-Ritense (approved):

ivo-ritense (approved):

PR #1751, opened 2025-03-11
story: zaaktypelink with case definition

mbritense:


zgw/catalogi-api/src/main/kotlin/com/ritense/catalogiapi/web/rest/CatalogiResource.kt
```diff
@@ -52,11 +53,11 @@ class CatalogiResource(
         return ResponseEntity.ok(zaakObjectTypes)
     }
 
-    @GetMapping("/v1/case-definition/{caseDefinitionName}/zaaktype/roltype")
+    @GetMapping("/v1/case-definition/{caseDefinitionId}/zaaktype/roltype")
```
ivo-ritense: When you need a case definition id you should use something like `/case-definition/{caseDefinitionKey}/version/{versionTag}` and construct the id in the method.

zgw/zaken-api/src/main/kotlin/com/ritense/zakenapi/domain/ZaakTypeLink.kt
```diff
@@ -45,10 +45,10 @@ data class ZaakTypeLink(
     @JsonProperty("id")
     val zaakTypeLinkId: ZaakTypeLinkId,
 
-    @Column(name = "document_definition_name", columnDefinition = "VARCHAR(50)", nullable = false)
+    @Column(name = "case_definition_id", columnDefinition = "VARCHAR(50)", nullable = false)
```
ivo-ritense: When linking to a case definition you don't need the @column annotation because it's already present in the CaseDefinitionId class. The lengtrh and notblank annotation i think are also not needed.

This does require you to add the column `case_definition_key` and `case_definition_version_tag` in the database.

app/gzac/src/main/java/com/ritense/gzac/listener/OpenZaakApplicationReadyEventListener.kt
ivo-ritense, line 1: I think this file gets to go. This was done to make a bootable app when checked out. Since we can now import this, we should replace this class by an import file.

app/gzac/build.gradle
```diff
@@ -10,9 +10,6 @@ dependencies {
 
     implementation("io.github.microutils:kotlin-logging")
 
-    compileOnly(project(":besluit"))
-    compileOnly(project(":openzaak"))
-    compileOnly(project(":contactmoment"))
     compileOnly(project(":objects-api"))
```
ivo-ritense: Shouldn't this module also be removed?

dependencies/valtimo-dependency-versions/build.gradle
```diff
@@ -71,16 +71,11 @@ dependencies {
...
 
         //GZAC
-        api(project(":besluit"))
-        api(project(":contactmoment"))
         api(project(":haalcentraal:haalcentraal-brp"))
-        api(project(":klant"))
         api(project(":plugin-authentication:notificaties-api-authentication"))
         api(project(":plugin-authentication:objecten-api-authentication"))
         api(project(":plugin-authentication:openzaak-plugin-authentication"))
         api(project(":objects-api"))
```
ivo-ritense: I think objects-api should also be removed

settings.gradle
```diff
@@ -58,7 +56,6 @@ include(
     "milestones",
     "notes",
     "objects-api",
```
ivo-ritense: I think this should be removed

zgw/zaken-api/src/main/resources/config/liquibase/changelog/20250312_zaak_type_link_to_case_definition.xml
```diff
@@ -19,9 +19,16 @@
                    xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
                    xsi:schemaLocation="http://www.liquibase.org/xml/ns/dbchangelog http://www.liquibase.org/xml/ns/dbchangelog/dbchangelog-3.8.xsd">
 
-    <changeSet id="1" author="Ivo Zaal">
```
ivo-ritense, line 22: Looks like you have overwritten an existing changeset, not sure if intended. If it was intended it should probably use a different id so it won't cause problems.

mbritense: If you look at this, it says I changed a file from openzaak, but this module is removed. Seems like git quirk to me, can you verify? I think I just added a new changelog file

ivo-ritense: I think you're right. Sometimes git tried to guess if files moved. This seems to be one of those cases.

ivo-ritense (approved):

PR #1752, opened 2025-03-11
Form flows for case definition

ivo-ritense:
Added link between case definition and form flows. Fixed some process link stuff that's connected to case definition change.


form-flow/src/main/kotlin/com/ritense/formflow/security/ValtimoFormFlowHttpSecurityConfigurer.kt
```diff
@@ -1,20 +1,20 @@
 /*
- *  Copyright 2015-2024 Ritense BV, the Netherlands.
```
marijnritense, line 2: ?

form-flow/src/main/kotlin/com/ritense/formflow/service/FormFlowDeploymentService.kt
```diff
@@ -124,13 +106,9 @@ class FormFlowDeploymentService(
...
     }
 
-    private fun loadFormFlowDefinitionsResources(): Array<Resource> {
-        return ResourcePatternUtils.getResourcePatternResolver(resourceLoader).getResources(FORM_FLOW_DEFINITIONS_PATH)
-    }
-
     companion object {
         private const val FORM_FLOW_SCHEMA_PATH = "classpath:config/form-flow/schema/formflow.schema.json"
-        private const val FORM_FLOW_DEFINITIONS_PATH = "classpath:config/form-flow/*.json"
+        private const val FORM_FLOW_DEFINITIONS_PATH = "classpath:config/case/*/*/form-flow/{formFlowKey}.json"
```
marijnritense, line 113: Tis is different from how we do form deployments (or any other for that matter). Can you make it more similar to those? And move the logic to the importer as well.

form-flow/src/main/kotlin/com/ritense/formflow/security/ValtimoFormFlowHttpSecurityConfigurer.kt
ivo-ritense: Looks like automatic reformat of the license because of extra whitespace.

marijnritense (approved):

form-flow/src/main/kotlin/com/ritense/formflow/service/FormFlowDeploymentService.kt
ivo-ritense: Not sure how it's different. This path is only used to later find if there's a source file for the form flow (to make it read only). We might want to do this differently later, but for now i'm not trying to change the functionality where it isn't needed.

PR #1754, opened 2025-03-13
feat: stop all running processes for a case

Klaas-Ritense:


Klaas-Ritense:
https://github.com/generiekzaakafhandelcomponent/gzac-issues/issues/139

David-Kooi-Ritense (dismissed):

process-document/src/main/kotlin/com/ritense/processdocument/service/ProcessDocumentsService.kt
```diff
@@ -41,6 +45,32 @@ class ProcessDocumentsService(
     private val repositoryService: RepositoryService,
     private val camundaRuntimeService: CamundaRuntimeService,
 ) {
+
+    @RunWithoutAuthorization
+    fun deleteAllProcessInstancesForThisDocument(execution: DelegateExecution, reason: String) {
```
marijnritense, line 47: This service isn't a process bean as far as I can tell. Maybe this method can be moved to a process bean instead, as that way, the `@RunWithoutAuthorization` also isn't necessary.

Klaas-Ritense: Hi @marijnritense, this service is a process bean. This can be seen here: [link](https://github.com/valtimo-platform/valtimo-backend-libraries/blob/next-minor/process-document/src/main/kotlin/com/ritense/processdocument/autoconfigure/ProcessDocumentsAutoConfiguration.kt#L141)

marijnritense: Then @RunWithoutAuthorization shouldn't be necessary I think.

ivo-ritense (dismissed):

ivo-ritense (approved):

mbritense (approved):

PR #1756, opened 2025-03-13
fix case deletion

Klaas-Ritense:


ivo-ritense (dismissed):

ThomasMinkeRitense:
@Klaas-Ritense Not yet mentioned, but yesterday I ran into an issue that I couldn't delete a case because there were notes attached to it. Would it be possible to look into that scenario as well?

Klaas-Ritense:
@ThomasMinkeRitense 
> Not yet mentioned, but yesterday I ran into an issue that I couldn't delete a case because there were notes attached to it. Would it be possible to look into that scenario as well?

That was already fixed here: https://github.com/valtimo-platform/valtimo-backend-libraries/pull/1724

ivo-ritense (approved):

marijnritense (approved):

PR #1771, opened 2025-03-27
Story/case management user endpoints

ivo-ritense:


form/src/main/kotlin/com/ritense/form/web/rest/FormResource.kt
```diff
@@ -70,9 +72,10 @@ class FormResource(
         @PathVariable formKey: String,
         @LoggableResource(resourceType = JsonSchemaDocument::class) @RequestParam(required = false) documentId: UUID?,
```
marijnritense, line 73: Can't you just make this required = true instead? I know it's a breaking change, but the code dictates the documentId always has to be present.

marijnritense (approved):

PR #1783, opened 2025-04-08
Bugfix: array in formio submission

Klaas-Ritense:


form/src/main/kotlin/com/ritense/form/service/impl/DefaultFormSubmissionService.kt
```diff
@@ -201,12 +205,34 @@ class DefaultFormSubmissionService(
...
-        val categorizedMap = formDefinition.inputFields
+        val targetKeysToJsonPointers = formDefinition.inputFields
             .filter { FormIoFormDefinition.NOT_IGNORED.test(it) }
             .filter { FormIoFormDefinition.isInputComponent(it) }
-            .mapNotNull { field ->
-                getTargetKeyValuePair(field, formData)
-            }.groupBy { (key, _) ->
+            .mapNotNull { field -> getTargetKeyJsonPointerPair(field, formData) }
+
+        val targetKeysToValues = targetKeysToJsonPointers
```
theo-ritense: This is pretty hard to read. Can you add some comments?

Klaas-Ritense: I've changed the approach. It should now be better readable

contract/src/main/java/com/ritense/valtimo/contract/json/patch/JsonPatchBuilder.java
```diff
@@ -93,28 +93,45 @@ public JsonPatchBuilder addJsonNodeValue(JsonNode destination, JsonPointer path,
...
-        String arrayPath = stringPath.substring(0, dashIndex);
-        for (int i = 0; ; i++) {
-            String testPath = arrayPath + "/" + i;
-            if (operations.stream().noneMatch(op -> op.getPath().equals(testPath))
-                && destination.at(testPath).isMissingNode()
-            ) {
-                String correctedPath = testPath + stringPath.substring(dashIndex + 2)
-                    .replace("/-", "/0");
-                return JsonPointer.compile(correctedPath);
+        } else if (dashIndex != -1) {
```
marijnritense, line 109: What if there's a - and + in there for some reason?

contract/src/main/java/com/ritense/valtimo/contract/json/patch/JsonPatchBuilder.java
```diff
@@ -93,28 +93,45 @@ public JsonPatchBuilder addJsonNodeValue(JsonNode destination, JsonPointer path,
     }
 
     /**
-     * This will adjust the first unindexed array position ('/-') to a new index depending on previous operations or destination object.
+     * This will adjust the first un-indexed array position ('/-' or '/+') to an index depending on previous operations or destination object.
+     * The '/-' will add the JSON to a new element in the array.
+     * The '/+' will append JSON to the last element in the array, or create a new element.
```
ivo-ritense: This line could be improved to say in which case a new element is created.

ivo-ritense (dismissed):

contract/src/main/java/com/ritense/valtimo/contract/json/patch/JsonPatchBuilder.java
```diff
@@ -93,28 +93,45 @@ public JsonPatchBuilder addJsonNodeValue(JsonNode destination, JsonPointer path,
...
+                String testPath = arrayPath + "/" + i;
+                if (operations.stream().noneMatch(op -> op.getPath().equals(testPath))
+                    && destination.at(testPath).isMissingNode()
+                ) {
+                    String correctedPath = testPath + stringPath.substring(dashIndex + 2)
+                        .replace("/-", "/0");
+                    return JsonPointer.compile(correctedPath);
+                }
+            }
+        } else {
```
marijnritense, line 121: You're missing the scenario where there is neither a - nor a + in the string. What happens then?

contract/src/main/java/com/ritense/valtimo/contract/json/patch/JsonPatchBuilder.java
```diff
@@ -93,28 +93,45 @@ public JsonPatchBuilder addJsonNodeValue(JsonNode destination, JsonPointer path,
...
+            String arrayPath = stringPath.substring(0, plusIndex);
```
marijnritense, line 122: If there is neither a - nor a plus, that means we get out of bounds exception

contract/src/main/java/com/ritense/valtimo/contract/json/patch/JsonPatchBuilder.java
Klaas-Ritense: Change comment

contract/src/main/java/com/ritense/valtimo/contract/json/patch/JsonPatchBuilder.java
Klaas-Ritense: Line 107 will handle when no - or + is in the string

contract/src/main/java/com/ritense/valtimo/contract/json/patch/JsonPatchBuilder.java
Klaas-Ritense: I've added a scenario when both the - and the + are present

marijnritense (dismissed):

anna-ritense:
@Klaas-Ritense the new `/+/` does not seem to take into account missing arrays. The Javadoc implies that a new index will be created if no items found in the array, but this is not the case when the array itself is missing. `/-/` had this fixed by Theo more than a year ago, we would expect `/+/` to behave the same.
image
From my input, I would expect the final patch path to be `/verzoeken/0/documenten/0/`

Klaas-Ritense:
Hi @anna-ritense , Ive added a test that verifies that the /+/ can add new arrays

marijnritense (approved):

ivo-ritense (approved):

PR #1784, opened 2025-04-08
Case definition test fixes

marijnritense:
Every module from A to O has been fixed.


importer/src/main/kotlin/com/ritense/importer/ValtimoImportService.kt
```diff
@@ -142,42 +144,62 @@ open class ValtimoImportService(
...
-//                    logger.debug { "Importing ${entry.fileName} with importer ${importer.type()}" }
-//                    importer.import(ImportRequest(entry.fileName, entry.content, caseDefinitionId))
-//                }
-//            }
-//        }
-//    }
-
     @Transactional
-    override fun import(inputStream: InputStream) {
+    override fun import(inputStream: InputStream, caseDefinitionIdList: List<CaseDefinitionId>) {
```
ivo-ritense, line 148: Isn't this very similar to the other method in this class. can't we combine them somehow?

ivo-ritense (approved):

PR #1789, opened 2025-04-10
#134476 Creating drafts

Klaas-Ritense:


case/src/main/java/com/ritense/document/service/DocumentDefinitionService.java
```diff
@@ -34,6 +32,8 @@ public interface DocumentDefinitionService {
 
     Page<? extends DocumentDefinition> findAll(Pageable pageable);
 
+    List<JsonSchemaDocumentDefinition> findAllBy(CaseDefinitionId caseDefinitionId);
```
ivo-ritense: Shouldn't this be the following because it's a generic interface?
```suggestion
    List<? extends DocumentDefinition> findAllBy(CaseDefinitionId caseDefinitionId);
```

case/src/main/kotlin/com/ritense/case/service/CaseDefinitionImporter.kt
```diff
@@ -52,14 +52,8 @@ class CaseDefinitionImporter(
...
 
-        if (existingCaseDefinition == null || forceDeploy) { // TODO: revisit forceDeploy this when doing drafts
-            val activeCaseDefinition = caseDefinitionRepository.findByActiveIsTrueAndIdKey(caseDefinition.id.key)
-            if (activeCaseDefinition == null || activeCaseDefinition.id.versionTag < caseDefinition.id.versionTag) {
-                caseDefinitionRepository.save(caseDefinition.copy(active = true))
-                activeCaseDefinition?.let {caseDefinitionRepository.save(it.copy(active = false)) }
-            } else {
-                caseDefinitionRepository.save(caseDefinition)
-            }
+        if (existingCaseDefinition == null) {
```
ivo-ritense: I think this might be too simple. The settings for example should still be changeable while the existing case definition is in draft.

Klaas-Ritense: @ivo-ritense I've fixed this in the 'CaseDefinitionDeploymentService'

case/src/main/java/com/ritense/document/service/DocumentDefinitionService.java
Klaas-Ritense: Changed

contract/src/main/kotlin/com/ritense/valtimo/contract/case_/CaseDefinitionId.kt
```diff
@@ -44,6 +44,8 @@ data class CaseDefinitionId(
             ?: throw IllegalArgumentException("Given version '$versionTag' is not a valid Semver version")
     )
 
+    constructor(caseDefinitionId: String) : this(caseDefinitionId.split(':')[0], caseDefinitionId.split(':')[2])
```
ivo-ritense, line 47: We might want to add a check to see if the array length is 2 so we can filter out some of the garbage input.

ivo-ritense (approved):

PR #1794, opened 2025-04-14
#136978 duplicating a draft

Klaas-Ritense:


process-document/src/main/kotlin/com/ritense/processdocument/listener/ProcessDefinitionAssociationCaseListener.kt
```diff
@@ -0,0 +1,63 @@
...
+) {
+
+    @RunWithoutAuthorization
+    @EventListener(CaseDefinitionCreatedEvent::class)
+    fun handleCaseDefinitionCreatedEvent(event: CaseDefinitionCreatedEvent) {
+        if (event.duplicate) {
+            service.findProcessDefinitionCaseDefinitions(event.basedOnCaseDefinitionId!!).forEach { association ->
+                service.createProcessDocumentDefinition(
+                    ProcessDocumentDefinitionRequest(
+                        processDefinitionId = association.id.processDefinitionId,
```
ivo-ritense: This is a tricky one. The process definition id would change as well so we can't straight up copy it.

Klaas-Ritense: I've solved this by combining the two listeners

PR #1804, opened 2025-04-17
feat: upgrade process management

mbritense:


core/src/main/java/com/ritense/valtimo/service/CamundaProcessService.java
```diff
@@ -319,6 +319,34 @@ public List<CamundaProcessDefinition> getDeployedDefinitions(CaseDefinitionId ca
         ));
     }
 
+    public List<CamundaProcessDefinition> getUnlinkedDeployedDefinitions() {
+        denyAuthorization();
+        return AuthorizationContext.runWithoutAuthorization(() ->
+            camundaRepositoryService.findProcessDefinitions(byActive(), Sort.by(NAME)).stream()
+                .filter(def -> def.getVersionTag() == null || !def.getVersionTag().startsWith("CD:"))
```
Klaas-Ritense, line 326: Making a SQL predicate for these, would be better. Something like: 
```
camundaRepositoryService.findProcessDefinitions(byActive().and(byUnlinked()).and(byLatestVersion(), Sort.by(NAME))
```

core/src/main/java/com/ritense/valtimo/service/CamundaProcessService.java
```diff
@@ -465,6 +502,49 @@ public void deploy(
         }
     }
 
+    @Transactional
+    public DeploymentWithDefinitions deployUnlinked(
```
Klaas-Ritense, line 506: This method seems to have a lot of code duplication. Can that be reduced?

PR #1839, opened 2025-05-15
Merge feat/draft system into feature/case-definition

Klaas-Ritense:
DONT SQUASH


case/src/main/kotlin/com/ritense/case/web/rest/dto/CaseDefinitionDraftCreateRequest.kt
```diff
@@ -0,0 +1,35 @@
...
+data class CaseDefinitionDraftCreateRequest(
+    val caseDefinitionKey: String,
+    val caseDefinitionVersion: String,
+    val name: String? = null,
+    val description: String? = null,
+    val basedOnCaseDefinitionVersion: String? = null,
+) {
+    fun getCaseDefinitionId(): CaseDefinitionId = CaseDefinitionId(caseDefinitionKey, caseDefinitionVersion)
+
+    fun getBasedOnCaseDefinitionId(): CaseDefinitionId? {
```
marijnritense, line 30: Since we still have time to change this: Should we rename this to `getReferenceCaseDefinitionId`? Or `getOriginalCaseDefinitionId`?

Klaas-Ritense: This is Git Branch terminology. I the future it might be possible to 'rebase' the case-definition on another case-definition.

ivo-ritense (approved):

PR #1856, opened 2025-05-20
Fixed deploying process definitions

ivo-ritense:


core/src/main/java/com/ritense/valtimo/service/CamundaProcessService.java
```diff
@@ -560,12 +546,33 @@ private boolean isProcessDefinitionPreviouslyDeployed(
...
+
+        var query = repositoryService
+            .createProcessDefinitionQuery()
+            .processDefinitionKey(processDefinitionKey)
+            .latestVersion()
+            .active();
+        if (caseDefinitionId != null) {
+            query = query.versionTag(CAMUNDA_CASE_DEFINITION_VERSION_TAG_PREFIX + caseDefinitionId);
+        } else {
+            query = query.withoutVersionTag();
```
Klaas-Ritense: tnh, I dont think this is correct since people are still allowed to use their own version tags. Is it possible to get it from the `bpmnModel`?

Klaas-Ritense (approved):

core/src/main/java/com/ritense/valtimo/service/CamundaProcessService.java
ivo-ritense: fixed

PR #1873, opened 2025-05-27
fix: form process links

mbritense:
Also fixes an issue where process links were not copied from the version the process def was based on, but the immediate previous version of the process def. Closes https://ritense.tpondemand.com/entity/139603-be-form-process-links-are-linked

### DIFF
--- a/core/src/main/kotlin/com/ritense/valtimo/camunda/domain/CamundaDeploymentSource.kt
+++ b/core/src/main/kotlin/com/ritense/valtimo/camunda/domain/CamundaDeploymentSource.kt
@@ -16,10 +16,39 @@
 
 package com.ritense.valtimo.camunda.domain
 
-enum class CamundaDeploymentSource(val value: String) {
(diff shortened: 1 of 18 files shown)


core/src/main/kotlin/com/ritense/valtimo/camunda/domain/CamundaDeploymentSource.kt
```diff
@@ -16,10 +16,39 @@
 
 package com.ritense.valtimo.camunda.domain
 
-enum class CamundaDeploymentSource(val value: String) {
-    SKIP_PROCESS_LINKS_COPY("SkipProcessLinksCopy");
-
+data class CamundaDeploymentSource @JvmOverloads constructor(
```
ivo-ritense: This might be an issue. The max size of the source column is 255 chars. The versiontag is up to 200 chars. Not sure this will fit.

Klaas-Ritense (approved):

marijnritense (approved):

PR #1881, opened 2025-06-02
Bugfix: Error when logged in without admin permissions

Klaas-Ritense:
Context: The frontend needs to have the endpoint below available for non-admins:
`GET /api/management/v1/case-definition/check`

### DIFF
--- a/case/src/main/kotlin/com/ritense/case/security/config/CaseHttpSecurityConfigurer.kt
+++ b/case/src/main/kotlin/com/ritense/case/security/config/CaseHttpSecurityConfigurer.kt
@@ -86,7 +86,7 @@ class CaseHttpSecurityConfigurer : HttpSecurityConfigurer {
                         )
                     ).hasAuthority(ADMIN)
                     .requestMatchers(antMatcher(POST, "/api/management/v1/case/import")).hasAuthority(ADMIN)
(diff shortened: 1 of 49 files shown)


case/src/main/kotlin/com/ritense/case/security/config/CaseHttpSecurityConfigurer.kt
```diff
@@ -86,7 +86,7 @@ class CaseHttpSecurityConfigurer : HttpSecurityConfigurer {
                         )
                     ).hasAuthority(ADMIN)
                     .requestMatchers(antMatcher(POST, "/api/management/v1/case/import")).hasAuthority(ADMIN)
-                    .requestMatchers(antMatcher(GET, "/api/management/v1/case-definition/check")).hasAuthority(ADMIN)
+                    .requestMatchers(antMatcher(GET, "/api/v1/case-definition/check")).authenticated()
```
marijnritense, line 89: Why is this change necessary? I can't quite figure out what the problem is based on the PR itself.

ivo-ritense (approved):

case/src/main/kotlin/com/ritense/case/security/config/CaseHttpSecurityConfigurer.kt
Klaas-Ritense: the endpoint is called for the case-list in the menu on the left. Admin can see cases that are in draft. Users cannot

PR #1899, opened 2025-06-19
Added event listener to copy process document links when a new draft is created

marijnritense:


process-document/src/main/kotlin/com/ritense/processdocument/listener/ProcessDocumentLinkEventListener.kt
```diff
@@ -0,0 +1,55 @@
...
+@Transactional
+@Component
+@SkipComponentScan
+class ProcessDocumentLinkEventListener(
+    private val caseDefinitionProcessLinkService: CaseDefinitionProcessLinkService
+) {
+
+    @RunWithoutAuthorization
+    @EventListener(CaseDefinitionCreatedEvent::class)
+    fun handleCaseDefinitionCreatedEvent(event: CaseDefinitionCreatedEvent) {
```
Klaas-Ritense, line 38: This listener is also needed here `@EventListener(CaseDefinitionPreDeleteEvent::class)`
Because when a draft is deleted, the process document link should also be deleted.

marijnritense: Fixed, thanks for pointing this out!

Klaas-Ritense (approved):

ivo-ritense (approved):

PR #1907, opened 2025-06-30
Feature/todos and deprecations

marijnritense:
Re enabled some tests and added more document table value resolver options:
- `documentDefinitionId` - This was previously `definitionId`.
- `documentDefinitionId.name` - This was previously `definitionId.name`
- `definitionId.key` - Refers to the case definition key.
- `definitionId.versionTag` - Refers to the case definition version tag.

These changes were made since the resolver works under the `case:` prefix.


process-document/src/main/kotlin/com/ritense/processdocument/resolver/DocumentTableValueResolver.kt
```diff
@@ -82,10 +82,10 @@ class DocumentTableValueResolver(
...
                 "createdBy" -> document.createdBy()
                 "createdOn" -> document.createdOn()
-                "definitionId" -> document.definitionId()
-                "definitionId.name" -> document.definitionId().name()
-                // TODO: change definitionId.caseDefinitionId
-                "definitionId.version" -> document.definitionId().caseDefinitionId()
+                "documentDefinitionId" -> document.definitionId()
+                "documentDefinitionId.name" -> document.definitionId().name()
+                "definitionId.key" -> document.definitionId().caseDefinitionId().key
+                "definitionId.versionTag" -> document.definitionId().caseDefinitionId().versionTag.version
```
Klaas-Ritense, line 88: Two different domains are now being mixed in a single value resolver. CaseDefinition and DocumentDefinition. It would be better to keep them separate

marijnritense: I agree it would be nicer to fully separate the two, but since a case definition and document definition are 1:1, I think it's fine here.

ivo-ritense: Not sure i agree. this isn't about getting the information of the document definition, it's about getting the information about the document. This includes relations like of what type the document is, and what the related case definition is.

ivo-ritense (dismissed):

process-document/src/main/kotlin/com/ritense/processdocument/resolver/DocumentTableValueResolver.kt
Klaas-Ritense: The inconsistenties:
- `case:createdOn` will _not_ get the created date from the case.
- `case:definitionId.key` _will_  get the definition key from the case.
- `DocumentTableValueResolver` suggests its _not_ about the case.
- `case:definitionId.versionTag` _will_  get the definition version from the case.

^ This suggests that two different domains are mixed.

marijnritense: Fair, then there has been a problem with this class since it was conceived, as it has had the `case` prefix since the beginning. This is misleading, as the data we're working with is document based.

On that note, this shows there's a flaw with the value resolvers. There can only ever be a single value resolver per prefix. But we see a big overlap in what data `DocumentJsonValueResolverFactory` and the `DocumentTableValueResolver` do. Where they differ is how they handle their values, and this is problematic since it would make sense for both of them to use the `doc` prefix.

ivo-ritense (approved):

Klaas-Ritense (approved):

PR #1938, opened 2025-08-13
story/more-specific-abonnementen-notificaties-api

marijnritense:
Made a separate trigger for the notificaties API module to create specific abonnementen whenever plugins are updated, deleted or created. 
Additionally, provides `valtimo.zgw.register-abonnementen` as a property to disallow abonnementen being registered, e.g. when working locally without the notificaties api.

Solves:
- https://github.com/generiekzaakafhandelcomponent/gzac-issues/issues/206
- https://github.com/generiekzaakafhandelcomponent/gzac-issues/issues/94


zgw/notificaties-api/src/main/kotlin/com/ritense/notificatiesapi/PluginsDeployedEventListener.kt
```diff
@@ -0,0 +1,188 @@
...
+                        knownNotificatiesApiAbonnementLinks,
+                        configurations
+                    )
+                    break
+                } catch (_: Exception) {
+                    --nrOfTriesLeftRetrievingAbonnementen
+                }
+            }
+            if (nrOfTriesLeftRetrievingAbonnementen == 0) {
+                val e = NotificatiesApiAbonnementException()
```
ivo-ritense: Would the original exception not be useful here as the cause?

zgw/notificaties-api/src/test/kotlin/com/ritense/notificatiesapi/PluginsDeployedEventListenerTest.kt
```diff
@@ -0,0 +1,87 @@
...
+        pluginsDeployedEventListener.setApplicationContext(context)
+    }
+
+    @Test
+    fun `should register nothing`() {
+        whenever(pluginService.getPluginConfigurations(any())).thenReturn(
+            emptyList()
+        )
+
+        pluginsDeployedEventListener.registerAbonnementenForNotificatiesApiPlugins()
```
ivo-ritense: While this is fine, i think it's better in cases like these to use `assertDoesNotThrow`. I had to look for a minute why this test did not have any assertions.

zgw/portaaltaak/src/main/kotlin/com/ritense/portaaltaak/PortaaltaakPlugin.kt
```diff
@@ -339,6 +341,38 @@ class PortaaltaakPlugin(
...
+            PluginConfigurationId
+                .existingId(objectManagement.objecttypenApiPluginConfigurationId)
+        ) as ObjecttypenApiPlugin
+
+
+        return listOf(
+            Abonnement.Kanaal(
+                naam = "objecten",
+                filters = mapOf(
+                    "objectType" to "${objecttypenApiPlugin.url}objecttypes/${objectManagement.objecttypeId}",
```
ivo-ritense, line 362: Would it not be better to move the construction of the url to the ObjecttypenApiPlugin? Reuseable and IMO a more logical place to put that responsibility.

zgw/verzoek/src/main/kotlin/com/ritense/verzoek/VerzoekPluginFactory.kt
```diff
@@ -18,16 +18,22 @@ package com.ritense.verzoek
...
 class VerzoekPluginFactory(
     pluginService: PluginService,
     private val caseDefinitionService: CaseDefinitionService,
     private val documentDefinitionService: JsonSchemaDocumentDefinitionService,
+    private val objectManagementService: ObjectManagementService
 ) : PluginFactory<VerzoekPlugin>(pluginService) {
 
     override fun create(): VerzoekPlugin {
-        return VerzoekPlugin(caseDefinitionService, documentDefinitionService)
+        return VerzoekPlugin(
```
ivo-ritense, line 33: The parameters here are inconsistent, is there a reason for that?

ivo-ritense:
I see no changes to the notificaties api plugin itself. shouldn't the automated abonnement creation be removed from that because it's now done by the listener?

zgw/notificaties-api/src/main/kotlin/com/ritense/notificatiesapi/PluginsDeployedEventListener.kt
marijnritense: fixed

zgw/notificaties-api/src/test/kotlin/com/ritense/notificatiesapi/PluginsDeployedEventListenerTest.kt
marijnritense: fixed

ivo-ritense (approved):

PR #1954, opened 2025-08-28
Personal stored quick search for case definitions

marijnritense:


case/src/main/kotlin/com/ritense/case/service/CaseInstanceService.kt
```diff
@@ -55,6 +64,75 @@ class CaseInstanceService(
...
+            )
+        )
+
+        require(
+            !quickSearchRepository.existsByCaseDefinitionKeyAndUserIdAndTitle(
+                caseDefinitionKey = caseDefinitionKey,
+                userId = currentUserId,
+                title = request.title
+            )
+        ) { "Failed to create quick save. A quick save for this user, for this case definition key, " +
```
ivo-ritense: You're using quick save while you're using quick search everywhere else.

case/src/main/kotlin/com/ritense/case/web/rest/CaseInstanceResource.kt
```diff
@@ -47,4 +51,40 @@ class CaseInstanceResource(
         val result = service.search(caseDefinitionName, searchRequest, pageable)
         return ResponseEntity.ok(result)
     }
+
+    @PostMapping("/v1/case/{caseDefinitionName}/stored-quick-search")
```
ivo-ritense: All instances of caseDefinitionName should be caseDefinitionKey

ivo-ritense (approved):

PR #1955, opened 2025-08-28
TP-143534: Made conditions generic + introduced condtitions to widgets

ivo-ritense:


ivo-ritense:
Moved to monorepo: https://github.com/valtimo-platform/valtimo/pull/12

PR #1959, opened 2025-09-02
TP-143543: Added hidden case list columns that can be configured by users individually.

marijnritense:


case/src/main/kotlin/com/ritense/case/web/rest/CaseDefinitionResource.kt
```diff
@@ -272,6 +274,30 @@ class CaseDefinitionResource(
         }
     }
 
+    @GetMapping("/v1/case/{caseDefinitionName}/hidden-list-column")
+    fun getHiddenCaseListColumnForUser(
+        @LoggableResource("documentDefinitionName") @PathVariable caseDefinitionName: String
+    ): ResponseEntity<List<CaseListColumnDto>> {
+        val currentUserId = UserManagementServiceHolder.currentInstance.currentUserId
```
ivo-ritense, line 281: Shouldn't we use the user email?

marijnritense: This is the implementation for currentUserId:

```
    @Override
    public String getCurrentUserId() {
        if (SecurityUtils.getCurrentUserAuthentication() != null) {
            return findUserRepresentationByEmail(SecurityUtils.getCurrentUserLogin()).orElseThrow(() ->
                new IllegalStateException("No user found for email: " + SecurityUtils.getCurrentUserLogin())
            ).getId();
        } else {
            return SYSTEM_ACCOUNT;
        }
    }
```
    
So I think this works.

ivo-ritense:
Moved to monorepo: https://github.com/valtimo-platform/valtimo/pull/21

